PR #23, opened 2018-10-12
update Dockerfile, with Giuseppe's newuidmap

AkihiroSuda:
EDIT: waiting for https://github.com/docker/cli/pull/1347 to be merged

- - -

With https://github.com/shadow-maint/shadow/pull/132 , we should no longer need `--privileged` for running RootlessKit inside Docker containers.

### DIFF
--- a/hack/test.sh
+++ b/hack/test.sh
@@ -3,6 +3,21 @@ set -eu -o pipefail -x
 
 cd $(dirname $0)/..
 
+ref=rootless-containers/rootlesskit:test
 # build -q but keep printing something (....) so as to avoid Travis timeout
-(while sleep 60; do echo -n .; done) & docker build -q -t rootless-containers/rootlesskit:test -f ./hack/test/Dockerfile . ; kill %1
-docker run -it --rm --privileged rootless-containers/rootlesskit:test
+(while sleep 60; do echo -n .; done) & docker build -q -t ${ref} -f ./hack/test/Dockerfile . ; kill %1
+
+# TODO: add `--security-opt unconfined-paths=/sys` when https://github.com/docker/cli/pull/1347 gets merged.
+# See https://github.com/rootless-containers/rootlesskit/pull/23 for the context.
+#
+# NOTE: bind-mounting /dev/net/tun is needed because we cannot mknod the tun device, at least with kernel < 4.18
+#
+# FIXME: --privilege is needed because the test suite compares rootlesskit with rootful `ip netns`.
+# TODO: split the rootful test to separate container and remove --privilege
+docker run \
+       --rm \
+       --security-opt seccomp=unconfined \
+       --security-opt apparmor=unconfined \
+       -v /dev/net/tun:/dev/net/tun \
+       --privileged \
+       ${ref}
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
`unshare -rmn mount -t sysfs none /sys` inside container seems to require some privilege..?

```console
$ id -u
1001
$ unshare -rmn mount -t sysfs none /sys
(success)
```

```console
$ docker run --rm --user 1000 --privileged ubuntu:18.04 unshare -rmn mount -t sysfs none /sys
(success)
```

```console
$ docker run --rm --user 1000 --security-opt seccomp=unconfined --security-opt apparmor=unconfined --cap-add all --net=host ubuntu:18.04 unshare -rmn mount -t sysfs none /sys
mount: /sys: permission denied.
```

AkihiroSuda:
@giuseppe @cyphar @jessfraz could you help me about the sysfs mount? ^^

AkihiroSuda:
Uh, read-only mount on `/sys` prohibits `unshare -rmn mount -t sysfs none /sys`...

https://github.com/moby/moby/blob/07ccc6d8c81d96e07f60184dbe03e75c552eba47/daemon/oci_linux.go#L607-L616

```console
$ docker run -it --privileged --rm --entrypoint bash rootless-containers/rootlesskit:test 
user@5ffda338c480:/$ unshare -rmn mount -t sysfs none /sys; echo $?
0
user@5ffda338c480:/$ sudo mount -o remount,ro /sys
user@5ffda338c480:/$ unshare -rmn mount -t sysfs none /sys; echo $?
mount: /sys: permission denied.
32
user@5ffda338c480:/$ mkdir /tmp/sys2
user@5ffda338c480:/$ unshare -rmn mount -t sysfs none /tmp/sys2; echo $?
mount: /tmp/sys2: permission denied.
32
user@5ffda338c480:/$ echo 'why :('
why :(
```

giuseppe:
> user@5ffda338c480:/$ echo 'why :(' why :(

I think the error happens because of: https://github.com/torvalds/linux/blob/master/fs/namespace.c#L3326-L3328

it should be fine to mount another `/sys` that is also 'ro'.  This works on Fedora with Linux 4.18.9-200.fc28.x86_64:

```
# docker run --rm  --security-opt seccomp=unconfined --cap-add all -ti ubuntu:18.04 \
  bash  -c 'umount /sys/firmware; \
  chroot --userspec 1000:1000 / unshare -rmn mount -o ro -t sysfs none /sys'
```

AkihiroSuda:
Thanks, but it seems we still need to unmount /sys/firmware as the root in the container...

giuseppe:
> Thanks, but it seems we still need to unmount /sys/firmware as the root in the container...

yes, there must already be a fully visible `/sys` mounted

cyphar:
Yes, this is quite similar to the other `mnt_already_visible` problems we had with `/proc`. Basically in order to mount any filesystem in a userns you need to have a non-masked mount of it in your mount namespace, with mount options that are not more restricted than what you're trying to mount.

(This is a somewhat odd restriction for `/proc` and `/sys` since they are namespaced implicitly -- and this has been brought to the attention of Eric in the past, but his proposed fix of `procfs2` is not something we should expect soon.)

AkihiroSuda:
updated PR.
Mounting sysfs is now skipped if it is not possible

PR #26, opened 2018-10-15
NodePort API spec & socat implementation

AkihiroSuda:
The API can be used for exposing TCP/UDP ports in the network namespace created along with the user namespace to the host network namespace.
    
* `api.sock` is created under the `rootlesskit --state-dir` directory
* API is described in `pkg/api/openapi.yaml`
  * `GET /v1/ports`
  * `POST /v1/ports`
   * `DELETE /v1/ports`
* `rootlessctl` CLI can be used as an API client:
  * `rootlessctl list-ports`
  * `rootlessctl add-ports`
  * `rootlessctl remove-ports`
* To expose child 80 as parent 8080: `rootlessctl --socket=/path/to/api.sock add-ports 0.0.0.0:8080:80/tcp`
    
Currently, only `socat` implementation is available as `rootlesskit --port-driver`.
The `slirp4netns`-native implementation is being experimented at https://github.com/rootless-containers/slirp4netns/pull/29 but it does not work with large MTUs at the moment.
    
The code under `pkg/port/socat` package might be also reusable for other projects.


pkg/api/client/client.go
```diff
@@ -0,0 +1,151 @@
...
+	lr := &io.LimitedReader{
+		R: r,
+		N: int64(maxBytes),
+	}
+	b, err := ioutil.ReadAll(lr)
+	if err != nil {
+		return b, err
+	}
+	if lr.N == 0 {
+		return b, errors.Errorf("expected at most %d bytes, got more (%d bytes are still left)", maxBytes, lr.N)
```
AkihiroSuda: TODO: fix error string

AkihiroSuda:
https://travis-ci.org/rootless-containers/rootlesskit/builds/445516404?utm_source=github_status&utm_medium=notification

slirp4netns mtu=65520

* outbound (child->slirp4netns->parent): 8.82 Gbits/sec  
* inbound (parent->socat->nsenter->socat->child): 8.82 Gbits/sec

not so bad

PR #28, opened 2018-10-29
unfork newuidmap/newgidmap

AkihiroSuda:

### DIFF
--- a/hack/test/Dockerfile
+++ b/hack/test/Dockerfile
@@ -10,12 +10,16 @@ CMD ["go","test","-v","-race","github.com/rootless-containers/rootlesskit/..."]
 FROM ubuntu:18.04 as build-c
 RUN apt update && apt install -y git make gcc automake autotools-dev libtool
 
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
It is really weird that this PR failed while https://github.com/moby/buildkit/pull/703 and https://github.com/genuinetools/img/pull/184 passed.. Maybe because this PR is based on Ubuntu rather than Alpine?

cc @giuseppe

AkihiroSuda:
The latest commit with `apt-get build-dep uidmap` should fix the issue

giuseppe:
> It is really weird that this PR failed while [moby/buildkit#703](https://github.com/moby/buildkit/pull/703) and [genuinetools/img#184](https://github.com/genuinetools/img/pull/184) passed.. Maybe because this PR is based on Ubuntu rather than Alpine?

might be that `<sys/capability.h>` is not present so the new feature is not compiled in

AkihiroSuda:
Probably yes but the previous commit already had `libcap-dev` ...

Issue #30, opened 2018-11-07
intermittent api.sock: bind: address already in use


AkihiroSuda:
```
+ rootlesskit --state-dir /run/user/1001/usernetes/rootlesskit --net=slirp4netns --mtu=65520 --copy-up=/etc --copy-up=/run --copy-up=/var/lib ./run.sh default-crio
[rootlesskit:parent] error: listen unix /run/user/1001/usernetes/rootlesskit/api.sock: bind: address already in use
```

u5surf:
Hi, I think it works better to be unlinking the unix domain socket file before listening
in pkg/parent/parent.go
https://github.com/rootless-containers/rootlesskit/blob/24449198253bb74f38d5c1a6270196f58f2871da/pkg/parent/parent.go#L247
ref:
https://troydhanson.github.io/network/Unix_domain_sockets.html

AkihiroSuda:
Thanks, if you don't mind could you open PR?

AkihiroSuda:
Fixed in #31

PR #31, opened 2018-11-08
unlink before bind #30

u5surf:
#30 
I fix unix domain socket is unlinked before net.Listen

### DIFF
--- a/pkg/parent/parent.go
+++ b/pkg/parent/parent.go
@@ -244,6 +244,10 @@ func listenServeAPI(socketPath string, backend *router.Backend) (apiCloser, erro
 	r := mux.NewRouter()
 	router.AddRoutes(r, backend)
 	srv := &http.Server{Handler: r}
+	err := os.RemoveAll(socketPath)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/parent/parent.go
```diff
@@ -244,6 +244,10 @@ func listenServeAPI(socketPath string, backend *router.Backend) (apiCloser, erro
 	r := mux.NewRouter()
 	router.AddRoutes(r, backend)
 	srv := &http.Server{Handler: r}
+	err := syscall.Unlink(socketPath)
+	if err != nil {
+		return nil, err
+	}
```
AkihiroSuda: Thanks, but CI seems failing

```
 rootlesskit --net=slirp4netns --mtu=1500 iperf3 -t 60 -c 10.0.2.2
[rootlesskit:parent] error: no such file or directory
```

Could you use `os.RemoveAll` instead of `syscall.Unlink`? https://golang.org/pkg/os/#RemoveAll

Also, please sign your commit with your full name?  (`git commit -a -s --amend`)

PR #53, opened 2019-03-23
use docker idtools for subid lookups

jaredbancroft:
runc/libcontainer/user lookups depend on users being local to the system and appearing in /etc/passwd and /etc/group. docker idtools fall back to using getent if local users aren't present which allows external lookups to work (systems using ldap with sssd). idtools looks up both subuid and subgid in a single map command which fails if a users primary group does not have the same name as their username, so username is passed as both arguments to idtools.NewIdentityMapping. os/user is used to determine uid and primary gid.


AkihiroSuda:
Thanks, could you sign your commit with your real name (`git commit -a -s --amend`) and also run `dep ensure`?

jaredbancroft:
All set

PR #55, opened 2019-03-27
move pcap from /tmp to XDG_RUNTIME_DIR

eschen42:
When working with https://get.dockerless.com/rootless I found that only one user could run the docker service at a time.  The reason is that the first to write /tmp/go.pcap was the owner of that file, and because other users could not write it (because of they did not own it), they could not start docker.  With this PR, as long as the `XDG_RUNTIME_DIR` environment variable is set, multiple users can run rootless docker, because each user's instance writes go.pcap to the runtime directory that they own.

This PR is intended to provoke discussion - I don't mind if it is rejected, but I would like suggestions for how to proceed.  The code that I changed is in fact from the moby/vpnkit package (though copied into this repository) - how would you like to manage this, assuming that you like this change?

### DIFF
--- a/vendor/github.com/moby/vpnkit/go/pkg/vmnet/vmnet.go
+++ b/vendor/github.com/moby/vpnkit/go/pkg/vmnet/vmnet.go
@@ -563,7 +563,14 @@ func (v *Vif) dhcp() (net.IP, error) {
 	ethernet := NewEthernetFrame(broadcastMAC, v.ClientMAC, 0x800)
 	ethernet.setData(ipv4.Bytes())
 
-	runDir := os.Getenv("XDG_RUNTIME_DIR")
+	// getEnv ref: https://stackoverflow.com/a/40326580
+	getEnv := func (key, fallback string) string {
+		if value, ok := os.LookupEnv(key); ok {
+			return value
+		}
+		return fallback
+	}
+	runDir := getEnv("XDG_RUNTIME_DIR", "/tmp")
 	pcapPath := fmt.Sprintf("%s/go.pcap", runDir)
 	file, err := os.Create(pcapPath)
 	if err != nil {
--- a/vendor/github.com/moby/vpnkit/go/pkg/vmnet/vmnet.go
+++ b/vendor/github.com/moby/vpnkit/go/pkg/vmnet/vmnet.go
@@ -564,7 +564,7 @@ func (v *Vif) dhcp() (net.IP, error) {
 	ethernet.setData(ipv4.Bytes())
 
 	// getEnv ref: https://stackoverflow.com/a/40326580
-	getEnv := func (key, fallback string) string {
+	getEnv := func(key, fallback string) string {
 		if value, ok := os.LookupEnv(key); ok {
 			return value
 		}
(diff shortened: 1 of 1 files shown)


vendor/github.com/moby/vpnkit/go/pkg/vmnet/vmnet.go
```diff
@@ -563,7 +563,9 @@ func (v *Vif) dhcp() (net.IP, error) {
 	ethernet := NewEthernetFrame(broadcastMAC, v.ClientMAC, 0x800)
 	ethernet.setData(ipv4.Bytes())
 
-	file, err := os.Create("/tmp/go.pcap")
+	runDir := os.Getenv("XDG_RUNTIME_DIR")
```
AkihiroSuda: not guaranteed to be set

vendor/github.com/moby/vpnkit/go/pkg/vmnet/vmnet.go
```diff
@@ -563,7 +563,9 @@ func (v *Vif) dhcp() (net.IP, error) {
...
+	pcapPath := fmt.Sprintf("%s/go.pcap", runDir)
```
AkihiroSuda: ioutil.TempFile?

AkihiroSuda:
thanks, but please open PR to https://github.com/moby/vpnkit/pulls

vendor/github.com/moby/vpnkit/go/pkg/vmnet/vmnet.go
eschen42: If set, this variable is set (for example) for user 1001 to /var/run/1001, which is properly owned by uid 1001.
This is true (at least when using rootlesskit to launch dockerd) whether running both outside the rootless context (when user.Uid is 1000) and inside (when user.Uid is 0).

Can you suggest another way to get real users' correct subdirectory of /run/user/ ?

vendor/github.com/moby/vpnkit/go/pkg/vmnet/vmnet.go
eschen42: From [https://golang.org/src/io/ioutil/tempfile.go](https://golang.org/src/io/ioutil/tempfile.go):
> It is the caller's responsibility to remove the file when no longer needed.

Perhaps it could be deleted in the Close method, but I don't know the exact purpose of the .pcap file.  Both in the original and in the PR, the same .pcap file is opened by the user UID outside the namespace and by UID 0 within the namespace - is the fact that both share the pcap file critical to the function of rootlesskit (I hope that it's obvious that I don't know what this file does).

vendor/github.com/moby/vpnkit/go/pkg/vmnet/vmnet.go
eschen42: One alternative is to fall back to the old behavior when the `XDG_RUNTIME_DIR` is not set.  

I am not a go programmer, but I have gotten the impression that, if the environment variable is not set, `runDir` will be set to an empty string (and I could test for that condition) - is that correct?

eschen42:
@AkihiroSuda I would be happy to open the PR against vpnkit, but first, ideally, I would prefer you please to see if you can give the advice that I requested in response to your comments.  Thank you for your time and advice.

AkihiroSuda:
wondering this pcap code was for debugging purpose and left over in the repo by accident?

@djs55

AkihiroSuda:
Thanks for contribution, but let me close this one

 #58

djs55:
Yes I think the pcap code was for debugging and left over by accident-- sorry, my mistake! @AkihiroSuda I've merged your moby/vpnkit PR to remove it.

PR #58, opened 2019-04-03
v0.3.0 (vendor vpnkit#472 #473)

### LINKED ISSUE #56: rootless docker works only for one user
When working with https://get.dockerless.com/rootless I found that only one user could run the docker service at a time.  The reason is that the first to write /tmp/go.pcap was the owner of that file, and because other users could not write it (because of they did not own it), they could not start docker.

I opened pull request #55 demonstrating a possible solution to this issue (which I implemented on my system to work around this issue).

AkihiroSuda:
https://github.com/moby/vpnkit/pull/472

Fix #56


AkihiroSuda:
DNM until https://github.com/moby/vpnkit/pull/473 gets merged

PR #59, opened 2019-04-04
new network driver: lxc-user-nic (experimental)

AkihiroSuda:
cc @giuseppe @cyphar


pkg/network/lxcusernic/lxcusernic.go
```diff
@@ -0,0 +1,156 @@
...
+func (d *parentDriver) MTU() int {
+	return d.mtu
+}
+
+func (d *parentDriver) ConfigureNetwork(childPID int, stateDir string) (*common.NetworkMessage, func() error, error) {
+	var cleanups []func() error
+	dummyLXCPath := "/dev/null"
+	dummyLXCName := "dummy"
+	dev := "eth0"
+	cmd := exec.Command(d.binary, "create", dummyLXCPath, dummyLXCName, strconv.Itoa(childPID), "veth", d.bridge, dev)
```
AkihiroSuda: @brauner @stgraber

Is it valid to call `lxc-user-net create $lxcpath $name $pid $type $bridge $nicname` with `$lxcpath="/dev/null"` `$name="dummy"`?

IIUC, `lxc-user-net` just ignores `$lxcpath` and `$name`, right? https://github.com/lxc/lxc/blob/master/src/lxc/cmd/lxc_user_nic.c

brauner: Yes, this should be valid. Let me add in a check that it is non-null so that we don't regress users in the future but calling it with a dummy path is fine.

AkihiroSuda: thanks!

AkihiroSuda:
cc @ibuildthecloud FYI

This eliminates the slirp4netns overhead completely (but needs SUID binary and setting up config file)

Issue #65, opened 2019-05-02
Support new pid namespace or ability to kill all children


ibuildthecloud:
This is a follow up to our discussion at DockerCon.

Rootlesskit (and the usernetes work) has been integrated it [k3s](http://github.com/rancher/k3s) but results in several usability issues.  For now it seems the simplest approach to making rootless with k3s more user friendly would be if rootkit could ensure some way that if the children of executed process died when the process died.  This could be done by creating a new pid namespace or possibly with cgroup (although I don't know if rootless can modify cgroups).

AkihiroSuda:
I'm just curious, why Pdeathsig doesn't work?

AkihiroSuda:
Minimal reproducer for the issue

```go
// Package main provides a reproducer for the PdeathSig issue.
//
// ```
// Terminal1$ ./pdeathsig-demo tail -f /dev/null
//
// Terminal2$ ps u
// ...
// suda       5705  0.0  0.0 102776  1844 pts/1    Sl+  02:47   0:00 ./pdeathsig-demo tail -f /dev/null
// suda       5709  0.0  0.0 102776  1520 pts/1    Sl+  02:47   0:00 /proc/self/exe tail -f /dev/null
// suda       5714  0.0  0.0   8068   820 pts/1    S+   02:47   0:00 tail -f /dev/null
// ...
// Terminal2$ kill 5705
// Terminal2$ ps u
// ...
// suda       5714  0.0  0.0   8068   820 pts/1    S    02:47   0:00 tail -f /dev/null
// ...
// ```
//
// The `tail -f /dev/null` process is expected to die, but actually it doesn't, due to some reason
// related to the newuidmap binary (SUID).
//
// When SKIP_NEWUIDMAP=1 is set, the `tail -f /dev/null` process dies as expected.
package main

import (
        "os"
        "os/exec"
        "strconv"
        "syscall"
)

func main() {
        if err := xmain(); err != nil {
                panic(err)
        }
}

func xmain() error {
        if os.Getenv("child") != "" {
                return child()
        }
        return parent()
}

func parent() error {
        cmd := exec.Command("/proc/self/exe", os.Args[1:]...)
        cmd.Env = append(os.Environ(), "child=1")
        cmd.Stdin = os.Stdin
        cmd.Stdout = os.Stdout
        cmd.Stderr = os.Stderr
        cmd.SysProcAttr = &syscall.SysProcAttr{
[...]

PR #66, opened 2019-05-14
child: reconfigure Pdeathsig (release v0.4.1)

AkihiroSuda:
The parent calls child with Pdeathsig, but it is cleared when newuidmap SUID binary is called
https://github.com/rootless-containers/rootlesskit/issues/65#issuecomment-492343646

Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>

### DIFF
--- a/pkg/child/child.go
+++ b/pkg/child/child.go
@@ -4,11 +4,13 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"runtime"
 	"strconv"
 	"syscall"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 
 	"github.com/rootless-containers/rootlesskit/pkg/common"
 	"github.com/rootless-containers/rootlesskit/pkg/copyup"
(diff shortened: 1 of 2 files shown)


pkg/child/child.go
```diff
@@ -187,6 +189,14 @@ func Child(opt Opt) error {
 	if msg.Stage != 1 {
 		return errors.Errorf("expected stage 1, got stage %d", msg.Stage)
 	}
+	// The parent calls child with Pdeathsig, but it is cleared when newuidmap SUID binary is called
+	// https://github.com/rootless-containers/rootlesskit/issues/65#issuecomment-492343646
+	runtime.LockOSThread()
+	err = unix.Prctl(unix.PR_SET_PDEATHSIG, uintptr(unix.SIGKILL), 0, 0, 0)
+	runtime.UnlockOSThread()
```
virtuald, line 196: This is incorrect, for PDEATHSIG the parent is the thread, not the process. I'll leave a more detailed comment on the original issue.

AkihiroSuda: Could you open an issue?

PR #96, opened 2020-01-06
Moved writing to child_pid file after child configuration

### LINKED ISSUE #37: "child_pid" file should not be created until the child is fully configured
Currently, `child_pid` is created under the state directory before the child is fully created.

This causes race for `nsenter -t $(cat child_pid)` scripts.

Zyqsempai:
Fixes #37 

Signed-off-by: Boris Popovschi <zyqsempai@mail.ru>

### DIFF
--- a/pkg/parent/parent.go
+++ b/pkg/parent/parent.go
@@ -149,12 +149,6 @@ func Parent(opt Opt) error {
 		}()
 	}
 
-	// after child is fully configured, write PID to child_pid file
-	childPIDPath := filepath.Join(opt.StateDir, StateFileChildPID)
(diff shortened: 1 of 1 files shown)


pkg/parent/parent.go
```diff
@@ -153,6 +149,12 @@ func Parent(opt Opt) error {
 		}()
 	}
 
+	// after child is fully configured, write PID to child_pid file
+	childPIDPath := filepath.Join(opt.StateDir, StateFileChildPID)
```
AkihiroSuda: thanks, but this should be after `portDriverInitComplete`

Also, probably the parent should remove the file before starting the child if the file already exists

Zyqsempai: This one is not enough https://github.com/rootless-containers/rootlesskit/blob/master/pkg/parent/parent.go#L73 ?

PR #99, opened 2020-01-22
Add reaper functionality to child

ibuildthecloud:
A new field Reaper is added to child.Opts to run a process
reaper when running the child target command.  This is needed
if you are running the child in a new PID namespace and you wish
to avoid zombies.


AkihiroSuda:
Thanks, will look into

AkihiroSuda:
How to test the reaper functionality?

Issue #97, opened 2020-01-07
support specifying mount propagation


AkihiroSuda:
Currently filesystems newly mounted on the host after RootlessKit has started cannot be seen from RootlessKit namespaces.

This could be solved by supporting non-private mount propagation.

Zyqsempai:
Oh, sounds good, I will work on it.

AkihiroSuda:
thanks

PR #98, opened 2020-01-07
Make base mount points shared

Zyqsempai:
Fixes #97 

Made base mount point shared.
Signed-off-by: Boris Popovschi <zyqsempai@mail.ru>

PR #109, opened 2020-02-25
[Carry 98] support --propagation=(rshared|rslave)

AkihiroSuda:
The default propagation is still `rprivate`.

`--propagation=rslave` is useful to propagate mounts from the host mount namespace.

`--propagation=rshared` doesn't work with `--copy-up` and isn't useful.

Fix https://github.com/rootless-containers/rootlesskit/issues/97
Carry #98


hack/test/integration.sh
```diff
@@ -0,0 +1,66 @@
...
+	*)
+		ERROR "Unknown propagation $propagation"
+		exit 1
+		;;
+	esac
+}
+
+test_propagation private
+test_propagation rprivate
+test_propagation slave
```
AkihiroSuda: fails inside container, but succeeds on host

PR #116, opened 2020-03-05
CLI: urfave/cli -> urfave/cli/v2 + docs change

AkihiroSuda:


AkihiroSuda:
@zhsj

https://salsa.debian.org/go-team/packages/rootlesskit/-/blob/74020a391d57ffc8d7154a534c0894d1057947ab/debian/patches/0001-Revert-CLI-urfave-cli-urfave-cli-v2.patch

What is the reason to revert this commit in Debian package? (Thanks for packaging :+1:)

zhsj:
> @zhsj
> 
> https://salsa.debian.org/go-team/packages/rootlesskit/-/blob/74020a391d57ffc8d7154a534c0894d1057947ab/debian/patches/0001-Revert-CLI-urfave-cli-urfave-cli-v2.patch
> 
> What is the reason to revert this commit in Debian package? (Thanks for packaging )

v2 is not available in Debian yet. https://packages.debian.org/sid/golang-github-urfave-cli-dev

zhsj:
to use v2 in Debian, I can

1. packaging v2 as a new package, which needs to go through the new queue, which is not very scale, https://ftp-master.debian.org/new.html moby/sys is currently in this queue.
2. update the old package to v2. but currently many packages use v1, so it's not the time yet.
3. patch rootlesskit, which is less effort.
4. keep v2 in vendor, which should be the last resort, as Debian policy don't want embedded source.

AkihiroSuda:
I prefer 4 (looks good) > 1 > 3 > 2 (looks bad) 😛

PR #122, opened 2020-03-07
print warning when preconditions are not satisfied + release v0.9.1

AkihiroSuda:
* print warning when unsupported propagation was specified (Suggested in https://github.com/moby/moby/pull/40635#issuecomment-595829139 , CC @thaJeztah @cpuguy83)
* print sysctl warnings
* print warning if running as root

### DIFF
--- a/pkg/parent/mount/mount_linux.go
+++ b/pkg/parent/mount/mount_linux.go
@@ -1,0 +1,82 @@
+// +build linux
+
+package mount
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+)
+
+const (
+	/* 36 35 98:0 /mnt1 /mnt2 rw,noatime master:1 - ext3 /dev/root rw,errors=continue
+	   (1)(2)(3)   (4)   (5)      (6)      (7)   (8) (9)   (10)         (11)
+
+	   (1) mount ID:  unique identifier of the mount (may be reused after umount)
+	   (2) parent ID:  ID of parent (or of self for the top of the mount tree)
+	   (3) major:minor:  value of st_dev for files on filesystem
+	   (4) root:  root of the mount within the filesystem
+	   (5) mount point:  mount point relative to the process's root
+	   (6) mount options:  per mount options
+	   (7) optional fields:  zero or more fields of the form "tag[:value]"
+	   (8) separator:  marks the end of the optional fields
+	   (9) filesystem type:  name of filesystem of the form "type[.subtype]"
+	   (10) mount source:  filesystem specific information or "none"
+	   (11) super options:  per super block options*/
+	mountinfoFormat = "%d %d %d:%d %s %s %s %s"
+)
+
+// Parse /proc/self/mountinfo because comparing Dev and ino does not work from
+// bind mounts
+func parseMountTable() ([]*Info, error) {
+	f, err := os.Open("/proc/self/mountinfo")
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	return parseInfoFile(f)
+}
+
+func parseInfoFile(r io.Reader) ([]*Info, error) {
+	var (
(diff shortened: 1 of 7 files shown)


pkg/parent/warn.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/rootless-containers/rootlesskit/pkg/parent/mount"
+)
+
+func warnPropagation(propagation string) {
+	mounts, err := mount.GetMounts()
+	if err != nil {
+		logrus.Warn("Failed to parse mountinfo")
```
thaJeztah: are we interested in the error as part of warnings? If so, perhaps `logrus.WithError(err).Warn(...`

pkg/parent/warn.go
```diff
@@ -0,0 +1,64 @@
+package parent
+
+import (
+	"io/ioutil"
+	"strconv"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
```
thaJeztah: is this repo using three groups (std, external, local) for imports? If not, (nit) remove this line

pkg/parent/warn.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// warnSysctl verifies /proc/sys/kernel/unprivileged_userns_clone and /proc/sys/user/max_user_namespaces
+func warnSysctl() {
+	uuc, err := ioutil.ReadFile("/proc/sys/kernel/unprivileged_userns_clone")
+	// The file may not exist on all distros
+	if err == nil {
+		i, err := strconv.ParseInt(strings.TrimSpace(string(uuc)), 10, 64)
+		if err == nil && i == 0 {
+			logrus.Warn("/proc/sys/kernel/unprivileged_userns_clone needs to be set to 1.")
```
thaJeztah: same question here (WithError?)

pkg/parent/mount/mount_linux.go
```diff
@@ -0,0 +1,82 @@
...
+	   (8) separator:  marks the end of the optional fields
+	   (9) filesystem type:  name of filesystem of the form "type[.subtype]"
+	   (10) mount source:  filesystem specific information or "none"
+	   (11) super options:  per super block options*/
+	mountinfoFormat = "%d %d %d:%d %s %s %s %s"
+)
+
+// Parse /proc/self/mountinfo because comparing Dev and ino does not work from
+// bind mounts
+func parseMountTable() ([]*Info, error) {
```
thaJeztah: Was this code copied from the moby "mount" pkg?

Should we make that package easier to import, you think? (separate repo or separate module if we're moving to go modules?)

AkihiroSuda: It is from runc: https://github.com/rootless-containers/rootlesskit/pull/122/files#diff-eb54b121747b544dfaf8954b8474c615R1

I think this amount of copypasta is fine. Separating the repo is rather likely to cause vendor hell.

pkg/parent/warn.go
AkihiroSuda: yes, three groups: https://github.com/rootless-containers/rootlesskit/blob/da1dfffa441bad0cfeeb35e7ca6423e5f9ee215d/pkg/parent/parent.go#L4-L26

pkg/parent/mount/mount_linux.go
thaJeztah: And, looks like runc copied it from moby 😓, so now it's in three locations https://github.com/moby/moby/blob/c4cf72bad325f90ebf021d7cb52c262992a687c7/pkg/mount/mountinfo_linux.go#L120-L130

pkg/parent/warn.go
AkihiroSuda: done

pkg/parent/warn.go
AkihiroSuda: done

PR #124, opened 2020-03-07
test: add lxc-user-nic benchmark

AkihiroSuda:


AkihiroSuda:
Weird, lxc-user-nic is slower than rootful veth (32.3 Gbits/sec  vs  41.8 Gbits/sec )
https://github.com/rootless-containers/rootlesskit/pull/124/checks?check_run_id=492490317

PR #131, opened 2020-03-17
Use moby/sys/mountinfo

kolyshkin:
*(this is a followup to https://github.com/rootless-containers/rootlesskit/pull/122)*

Drop the implementation forked from runc, use moby/sys/mountinfo.

Use SingleEntryFilter as we're only interested in / mount.

NOTE we're no longer checking for m.Root == "/", only for
the mount point.


kolyshkin:
@AkihiroSuda @thaJeztah PTAL

kolyshkin:
Can't figure out why CI fails :(

pkg/parent/warn.go
```diff
@@ -5,27 +5,17 @@ import (
...
-	for _, m := range mounts {
-		if m.Root == "/" && m.Mountpoint == "/" {
-			root = m
-		}
-	}
-	if root == nil {
-		logrus.Warn("Failed to parse mountinfo of the root filesystem")
-		return
-	}
+	root = mounts[0]
```
AkihiroSuda:
```
# github.com/rootless-containers/rootlesskit/pkg/parent
src/github.com/rootless-containers/rootlesskit/pkg/parent/warn.go:18:2: undefined: root
src/github.com/rootless-containers/rootlesskit/pkg/parent/warn.go:23:68: undefined: root
src/github.com/rootless-containers/rootlesskit/pkg/parent/warn.go:25:4: undefined: root
```

AkihiroSuda:
Thanks, LGTM except the compilation failure

kolyshkin:
oops, fixed

Issue #185, opened 2020-10-27
Default path of StateDir should be locked for preventing automatic clean-up by systemd-tmpfiles.


haruo31:
rootlesskit creates `/tmp/rootlesskitXXXX` directory if rootlesskit is launched without `--state-dir` option. But files in StateDir have no protection for automatic cleanup from systemd-tmpfiles. It could be lost `*.sock` files unexpectedly.

the systemd documentation describes how to protect files from automatic clean-up. (I suppose the number 3 is suitable for rootlesskit.)
https://systemd.io/TEMPORARY_DIRECTORIES/

AkihiroSuda:
Thank you for reporting. The number 3 SGTM. Would you be interested in opening a PR?

AkihiroSuda:
Fixed in #188, thanks @HappyTobi

PR #188, opened 2020-11-03
Lock StateDir

HappyTobi:
Hi all,

here is the FIX for the created issue #185 

## Description

If the user don't pass the `--state-dir` argument the internal `StateDirTemp` flag will be set.
On running the `Parent` function a Shared BSD Lock  [LOCK_SH](https://man7.org/linux/man-pages/man2/flock.2.html) will be set on the stateDir.

All locks will be removed with the `Rootlesskit` `StateFileLock` (`/tmp/rootlesskitXXX/lock`).

I also added a test to check if the lock was created, but to test that on a real system you have to read the verify part.

## Verify Systemd-tmpfiles cleanup resilience 

OS:  Ubuntu 18.04 LTS

1. Check tmp settings
`cat /usr/lib/tmpfiles.d/tmp.conf`

2. Change them if no cleanup for "old files" was configured.
`sudo nano /usr/lib/tmpfiles.d/tmp.conf`
See possible configuration settings: [tmpfiles.d(5)](https://man7.org/linux/man-pages/man5/tmpfiles.d.5.html)
**Example:** (for testing):
```D /tmp 1777 root root 10s -```
**Now all files will be deleted that are older than 10sec.**

3. Run systemd-tmpfiles cleanup manually 
To see whats happen we run the job with debug logs
`sudo env SYSTEMD_LOG_LEVEL=debug systemd-tmpfiles --clean` 

4. Check stateDir
The `rootlessXXX` stateDir in `/tmp/` still exists.

5. Show Lock
`lsof /tmp/rotlesskitXXXX`

 
Bg 
Tobi

### DIFF
--- a/pkg/parent/parent.go
+++ b/pkg/parent/parent.go
@@ -32,6 +32,7 @@ import (
 type Opt struct {
 	PipeFDEnvKey     string               // needs to be set
 	StateDir         string               // directory needs to be precreated
+	StateDirTemp     bool                 // set flag if StateDir was created in /tmp/
 	StateDirEnvKey   string               // optional env key to propagate StateDir value
 	NetworkDriver    network.ParentDriver // nil for HostNetwork
 	PortDriver       port.ParentDriver    // nil for --port-driver=none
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by AkihiroSuda** [comment]
Thanks, could you sign off the commit? (`git commit -a -s --amend`)

File: pkg/parent/parent.go [hunk 1]
```diff
@@ -77,12 +78,33 @@ func checkPreflight(opt Opt) error {
 	return nil
 }
 
+func createCleanupLock(sDir string, tempDir bool) error {
+	//lock state dir when using /tmp/ path
```
AkihiroSuda, line 82: Does something bad happen if we always lock the dir regardless to whether it is under /tmp or not?

pkg/parent/parent.go
```diff
@@ -77,12 +78,33 @@ func checkPreflight(opt Opt) error {
...
+	if tempDir {
+		stateDir, err := os.Open(sDir)
+		if err != nil {
+			return err
+		}
+		err = unix.Flock(int(stateDir.Fd()), unix.LOCK_SH)
+		if err != nil {
+			return errors.Errorf("failed to lock directory %s", sDir)
```
AkihiroSuda, line 90: can be just a warning rather than a critical error?

pkg/parent/parent.go
HappyTobi: nope that should be no problem.
I only skipped that because you didn't need that, but it's simpler and better to understand.

pkg/parent/parent.go
HappyTobi: yes for sure I will change that

AkihiroSuda (approved):
Thanks!

PR #192, opened 2020-11-11
Fix up privileged port exposure

rhatdan:
This error is showing up looking like

podman run -p 80:80 fedora echo hello
Error: failed to expose ports via rootlessport: "cannot expose privileged port 80, you might need to add \"net.ipv4.ip_unprivileged_port_start=0\" (currently 1024) to /etc/sysctl.conf, or choose a larger port number (>= 1024): listen tcp 0.0.0.0:80: bind: permission denied\n"

When run in Podman, trying to clean it up a little so the special chars \" to
not show up, and I think it makes more senset to tell users to start have priv
port, at least from a security point of view.  Users probably want to keep this above port 22.

net.ipv4.ip_unprivileged_port_start=80

Rather then
net.ipv4.ip_unprivileged_port_start=0

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

### DIFF
--- a/pkg/port/builtin/parent/parent.go
+++ b/pkg/port/builtin/parent/parent.go
@@ -111,7 +111,7 @@ func annotateEPERM(origErr error, spec port.Spec) error {
 		// origErr is unrelated to ip_unprivileged_port_start
 		return origErr
 	}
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thank you!

AkihiroSuda:
Released v0.11.1 with this

PR #206, opened 2021-01-20
port: add ChildIP

giuseppe:
allow users to override the IP to use for the connection inside the
network namespace.

It is useful e.g. with slirp4netns to override the IP to "10.0.2.100".

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/pkg/port/builtin/child/child.go
+++ b/pkg/port/builtin/child/child.go
@@ -106,7 +106,21 @@ func (d *childDriver) handleConnectRequest(c *net.UnixConn, req *msg.Request) er
 		return errors.Errorf("unknown proto: %q", req.Proto)
 	}
 	var dialer net.Dialer
-	targetConn, err := dialer.Dial(req.Proto, fmt.Sprintf("127.0.0.1:%d", req.Port))
+	ip := req.IP
+	if ip == "" {
+		ip = "127.0.0.1"
+	} else {
+		p := net.ParseIP(ip)
+		if p == nil {
+			return errors.Errorf("invalid IP: %q", ip)
+		}
+		p = p.To4()
+		if p == nil {
+			return errors.Errorf("unsupported IP (v6?): %s", ip)
+		}
+		ip = p.String()
+	}
+	targetConn, err := dialer.Dial(req.Proto, fmt.Sprintf("%s:%d", ip, req.Port))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


giuseppe:
it is needed for Podman: https://github.com/containers/podman/issues/5138

AkihiroSuda:
> it is needed for Podman: [containers/podman#5138](https://github.com/containers/podman/issues/5138)

This doesn't seem to be able to propagate src IP. Wouldn't this just make the src IP to 10.0.2.100 ?

giuseppe:
Yes, it won't propagate the srcIP but at least we can set it to something different than 127.0.0.1 that causes issues with some container images

giuseppe:
@AkihiroSuda are you fine with the change as it is?

pkg/port/builtin/child/child.go
```diff
@@ -106,7 +106,11 @@ func (d *childDriver) handleConnectRequest(c *net.UnixConn, req *msg.Request) er
 		return errors.Errorf("unknown proto: %q", req.Proto)
 	}
 	var dialer net.Dialer
-	targetConn, err := dialer.Dial(req.Proto, fmt.Sprintf("127.0.0.1:%d", req.Port))
+	ip := req.IP
```
AkihiroSuda, line 109: We need to verify that this is a valid IPv4 string

pkg/port/builtin/child/child.go
```diff
@@ -106,7 +106,11 @@ func (d *childDriver) handleConnectRequest(c *net.UnixConn, req *msg.Request) er
...
+	if ip == "" {
+		ip = "127.0.0.1"
+	}
+	targetConn, err := dialer.Dial(req.Proto, fmt.Sprintf("%s:%d", ip, req.Port))
```
AkihiroSuda, line 113: Could you also update pkg/port/{slirp4netns,socat}?

pkg/port/port.go
```diff
@@ -10,6 +10,7 @@ type Spec struct {
 	ParentIP   string `json:"parentIP,omitempty"` // IPv4 address. can be empty (0.0.0.0).
 	ParentPort int    `json:"parentPort,omitempty"`
 	ChildPort  int    `json:"childPort,omitempty"`
+	ChildIP    string `json:"childIP,omitempty"`
```
AkihiroSuda, line 13: Please add godoc comment to clarify that this defaults to 127.0.0.1

AkihiroSuda (commented):
Looks good but a couple of nits. Thanks

AkihiroSuda (approved):
Thanks, I'll ship a new release tomorrow

PR #213, opened 2021-02-11
Refactor ParsePortSpec to handle IPv6 addresses, and improve validation

thaJeztah:
Refactor of portutil, related to https://github.com/moby/moby/pull/41908 and https://github.com/rootless-containers/rootlesskit/pull/207

### Refactor ParsePortSpec to handle IPv6 addresses

The PortSpec uses colons a delimiter between parts. If an IPv6 address is used, this leads to the colons in the IP-address to be handled as delimiter.

Commit 1d78bed8423d3377e04c4bdd1e1a045fb42e2942 (https://github.com/rootless-containers/rootlesskit/pull/207) also added support for "ChildIP" to the portspec format, which (adding to the above) made parsing the format even more ambiguous.

This patch rewrites ParsePortSpec to handle IPv6 addresses, but requires those addresses to be wrapped by square brackes.

### ValidatePortSpec: accept sctp as valid proto

There was an inconsistency between ParsePortSpec and ValidatePortSpec; ParsePortSpec accepted "sctp" as a valid proto, whereas ValidatePortSpec did not.

This patch re-uses the `validateProto()` function to validate.

### ValidatePortSpec: validate ChildIP (if set)

The function already validated ParentIP, but did not validate ChildIP; also added a minimal test

### Use net.JoinHostPort() to join IP-addresses with port

This makes sure the IPv6 addresses are wrapped with square brackets (e.g., `[::1]:8080`)

### DIFF
--- a/pkg/port/socat/socat.go
+++ b/pkg/port/socat/socat.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"strconv"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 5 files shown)


AkihiroSuda:
Thanks!

thaJeztah:
Testing it on the moby PR to see if there's things missing still, but I _think_ remaining failures are not in this code (will have to check though)

thaJeztah:
Moved out of draft; tried this branch separate from the libnetwork changes, and rootless CI on moby passes with this; https://github.com/moby/moby/pull/42021

Well can look into the failures with the libnetwork update separately

pkg/port/socat/socat.go
```diff
@@ -149,21 +150,23 @@ func createSocatCmd(ctx context.Context, spec port.Spec, logWriter io.Writer, ch
 		}
 		ip = p.String()
 	}
-	var cmd *exec.Cmd
+	var (
+		cmd   *exec.Cmd
+		proto string
+		hp    = net.JoinHostPort(ip, strconv.Itoa(spec.ChildPort))
+	)
 	switch spec.Proto {
```
AkihiroSuda, line 158: should have `default` (not new in this PR)

pkg/port/socat/socat.go
```diff
@@ -149,21 +150,23 @@ func createSocatCmd(ctx context.Context, spec port.Spec, logWriter io.Writer, ch
...
-			fmt.Sprintf("UDP-LISTEN:%d,bind=%s,reuseaddr,fork,rcvbuf=65536,sndbuf=65536", spec.ParentPort, ipStr),
-			fmt.Sprintf("EXEC:\"%s\",nofork",
-				fmt.Sprintf("nsenter -U -n --preserve-credentials -t %d socat STDIN UDP4:%s:%d", childPID, ip, spec.ChildPort)))
+		proto = "UDP"
 	}
+	cmd = exec.CommandContext(ctx,
+		"socat",
+		fmt.Sprintf("%s-LISTEN:%d,bind=%s,reuseaddr,fork,rcvbuf=65536,sndbuf=65536", proto, spec.ParentPort, ipStr),
+		fmt.Sprintf(`EXEC:"%s",nofork`,
+			fmt.Sprintf("nsenter -U -n --preserve-credentials -t %d socat STDIN %s4:%s", childPID, proto, hp)))
```
AkihiroSuda, line 168: I don't understand why this sprintf is nested, but not new in this PR 🤷‍♂️

pkg/port/socat/socat.go
```diff
@@ -149,21 +150,23 @@ func createSocatCmd(ctx context.Context, spec port.Spec, logWriter io.Writer, ch
 		}
 		ip = p.String()
 	}
-	var cmd *exec.Cmd
+	var (
+		cmd   *exec.Cmd
+		proto string
+		hp    = net.JoinHostPort(ip, strconv.Itoa(spec.ChildPort))
```
AkihiroSuda, line 156: should error out for IPv6

AkihiroSuda (approved):
Thanks

pkg/port/socat/socat.go
thaJeztah: Oh! I was also looking at that; at first thought there was a reason, but didn't see any either. Should've changed it while I was modifying the code; let me do a follow-up

thaJeztah: https://github.com/rootless-containers/rootlesskit/pull/215

PR #223, opened 2021-02-22
docker-proxy: don't discard port removal errors

thaJeztah:
related to https://github.com/moby/moby/pull/41908#issuecomment-783332755

### DIFF
--- a/cmd/rootlesskit-docker-proxy/main.go
+++ b/cmd/rootlesskit-docker-proxy/main.go
@@ -35,7 +35,7 @@ func main() {
 	}
 }
 
-func xmain(f *os.File) error {
+func xmain(f *os.File) (retErr error) {
 	containerIP := flag.String("container-ip", "", "container ip")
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):

thaJeztah:
I should add that testing it on https://github.com/moby/moby/pull/41908#issuecomment-783332755 didn't show errors, so the problem I'm looking into probably isn't related to that.

I guess this change "won't hurt" to have (as it would show errors (if any)). I should also mention that I couldn't come up with a way to actually test this (I _think_ it works, but if you have a clever way to test it, let me know)

thaJeztah:
Closing; this is no longer needed, and now implemented through; https://github.com/rootless-containers/rootlesskit/blob/be5b9011d908d0906243bcf80f2f58ab90156fbd/cmd/rootlesskit-docker-proxy/main.go#L111-L117

PR #249, opened 2021-04-14
rootlesskit-docker-proxy: exit without error when port driver does not support v6 (fix Docker 20.10.6 + slirp4netns port driver)

### LINKED ISSUE #250: [regression in Docker 20.10.6] slirp4netns port driver fails: "Timed out proxy starting the userland proxy."
Rootless Docker  20.10.6 + RootlessKit v0.14.1 + slirp4netns port driver fails

```console
$ cat ~/.config/systemd/user/docker.service.d/override.conf 
[Service]
Environment=DOCKERD_ROOTLESS_ROOTLESSKIT_PORT_DRIVER="slirp4netns"

$ docker --context=rootless run --rm -p 8080:80 nginx:alpine
docker: Error response from daemon: driver failed programming external connectivity on endpoint dreamy_gauss (93092ec62bc18d4190f18c05f188505c11acce4feef7ea2eb259805c18edfd85): Timed out proxy starting the userland proxy.
```

The builtin port driver with slirp4netns network driver works.

The both drivers were working with Docker v20.10.5.

AkihiroSuda:
Fix #250 (regression in Docker 20.10.6)

The error only affects slirp4netns port driver users (non-default), not slirp4netns network driver users.


AkihiroSuda:
Rootless Docker + slirp4netns port forwarder seems failing.

```
docker: Error response from daemon: driver failed programming external connectivity on endpoint nginx (2d7a1e85c79ab1a6ac65bff78855fd87edc0a9a7c050c24a2eae999623bcb599): Timed out proxy starting the userland proxy.
Error: Process completed with exit code 125.
```

builtin port driver is working.

The both drivers were working with Docker v20.10.5.

AkihiroSuda:
cc @thaJeztah

thaJeztah:
@AkihiroSuda in this case, is it that the _driver_ doesn't support IPv6, or the _host_? While this (generally) seems ok to do, if it's the latter, I'm wondering if the docker daemon / libnetwork failed to detect if IPv6 _can_ be used (and should not have attempted to publish on IPv6 in the first place)

AkihiroSuda:
> in this case, is it that the driver doesn't support IPv6, or the host? 

The driver.
More accurately, whether the driver accepts "tcp6"/"udp6" as the protocol string.

thaJeztah:
Ah, thanks, so handling of dockerd is correct otherwise (and in the "non-rootless" case)?

AkihiroSuda:
Yes

PR #262, opened 2021-08-02
port/builtin: RemovePort() block until conn is closed

Luap99:
RemovePort() should block until the connection is closed. This allows
the caller to safely use AddPort() afterwards.

Signed-off-by: Paul Holzinger <pholzing@redhat.com>

### DIFF
--- a/pkg/port/builtin/parent/parent.go
+++ b/pkg/port/builtin/parent/parent.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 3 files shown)


AkihiroSuda (approved):
Thanks, will release v0.14.3 soon with this

AkihiroSuda:
Seems to caused regression https://github.com/containerd/nerdctl/pull/306#issuecomment-892388898

pkg/port/builtin/parent/parent.go
```diff
@@ -140,8 +141,13 @@ func (d *driver) AddPort(ctx context.Context, spec port.Spec) (*port.Status, err
 	}
 	routineStopCh := make(chan struct{})
 	routineStop := func() error {
-		close(routineStopCh)
-		return nil // FIXME
+		routineStopCh <- struct{}{}
```
AkihiroSuda, line 144: This can be delivered to the `copyConnToChild(c, socketPath, spec, stopCh)` routine in `tcp.go`, so this does not work as expected

AkihiroSuda: Opened PR https://github.com/rootless-containers/rootlesskit/pull/265 to  fix this

PR #265, opened 2021-08-04
port/builtin: RemovePort() block until conn is closed, take 2 + Release v0.14.4

AkihiroSuda:
Revert #262 (https://github.com/rootless-containers/rootlesskit/pull/262/files#r682338302) and add an alternative.


Luap99:
Thanks @AkihiroSuda, sorry for missing that.

PR #319, opened 2022-07-26
Fix ipv6 feature error typo

kowalski7cc:
The error was incorreclty telling "this version of slirp4netns does not support --enable-sandbox" instead of "this version of slirp4netns does not support --enable-ipv6"


AkihiroSuda (approved):
Thanks

PR #358, opened 2023-03-07
new network driver: `pasta` (with port driver `implicit`)

AkihiroSuda:
Pasta: https://passt.top/passt/
Usage: `rootlesskit --net=pasta --port-driver=implicit`

- No support for explicit port forwarding (`rootlessctl add-ports`),
  as pasta doesn't support it yet.
  Use `--port-driver=implicit` to let pasta forward TCP ports implicitly.
  The forwarded ports are not visible in `rootlessctl list-ports`.

- No support for forwarding UDP ports

- Tested with pasta 2023_06_25.32660ce on Ubuntu 23.04.
  Doesn't work with 2023_06_03.429e1a7: `Option --no-copy-routes needs --config-net`
  (This is printed despite that `--no-copy-routes` is not specified)

- Doesn't work with Ubuntu 23.04's dpkg (passt_0.0~git20230216.4663ccc-1_amd64.deb):
  `Couldn't open user namespace /proc/51813/ns/user: Permission denied`
  Likely to be related to AppArmor.
  `sudo apparmor_parser -R /etc/apparmor.d/usr.bin.passt` can eliminate this error, but pasta still fails with another error ( `Couldn't get any nameserver address`)

### DIFF
--- a/cmd/rootlesskit/main.go
+++ b/cmd/rootlesskit/main.go
@@ -17,6 +17,7 @@ import (
 	"github.com/rootless-containers/rootlesskit/pkg/common"
 	"github.com/rootless-containers/rootlesskit/pkg/copyup/tmpfssymlink"
 	"github.com/rootless-containers/rootlesskit/pkg/network/lxcusernic"
+	"github.com/rootless-containers/rootlesskit/pkg/network/pasta"
 	"github.com/rootless-containers/rootlesskit/pkg/network/slirp4netns"
 	"github.com/rootless-containers/rootlesskit/pkg/network/vpnkit"
 	"github.com/rootless-containers/rootlesskit/pkg/parent"
@@ -77,9 +78,14 @@ See https://rootlesscontaine.rs/getting-started/common/ .
 		}, CategoryState),
 		Categorize(&cli.StringFlag{
 			Name:  "net",
-			Usage: "network driver [host, slirp4netns, vpnkit, lxc-user-nic(experimental)]",
+			Usage: "network driver [host, pasta(experimental), slirp4netns, vpnkit, lxc-user-nic(experimental)]",
 			Value: "host",
 		}, CategoryNetwork),
+		Categorize(&cli.StringFlag{
+			Name:  "pasta-binary",
+			Usage: "path of pasta binary for --net=pasta",
+			Value: "pasta",
+		}, CategoryPasta),
 		Categorize(&cli.StringFlag{
 			Name:  "slirp4netns-binary",
 			Usage: "path of slirp4netns binary for --net=slirp4netns",
@@ -112,16 +118,16 @@ See https://rootlesscontaine.rs/getting-started/common/ .
 		}, CategoryLXCUserNic),
 		Categorize(&cli.IntFlag{
 			Name:  "mtu",
-			Usage: "MTU for non-host network (default: 65520 for slirp4netns, 1500 for others)",
+			Usage: "MTU for non-host network (default: 65520 for pasta and slirp4netns, 1500 for others)",
 			Value: 0, // resolved into 65520 for slirp4netns, 1500 for others
 		}, CategoryNetwork),
 		Categorize(&cli.StringFlag{
 			Name:  "cidr",
-			Usage: "CIDR for slirp4netns network (default: 10.0.2.0/24)",
+			Usage: "CIDR for pasta and slirp4netns networks (default: 10.0.2.0/24)",
 		}, CategoryNetwork),
 		Categorize(&cli.StringFlag{
 			Name:  "ifname",
-			Usage: "Network interface name (default: tap0 for slirp4netns and vpnkit, eth0 for lxc-user-nic)",
+			Usage: "Network interface name (default: tap0 for pasta, slirp4netns, and vpnkit; eth0 for lxc-user-nic)",
 		}, CategoryNetwork),
 		Categorize(&cli.BoolFlag{
 			Name:  "disable-host-loopback",
@@ -129,7 +135,7 @@ See https://rootlesscontaine.rs/getting-started/common/ .
 		}, CategoryNetwork),
 		Categorize(&cli.BoolFlag{
 			Name:  "ipv6",
-			Usage: "enable IPv6 routing. Unrelated to port forwarding. Only supported for slirp4netns. (experimental)",
+			Usage: "enable IPv6 routing. Unrelated to port forwarding. Only supported for pasta and slirp4netns. (experimental)",
 		}, CategoryNetwork),
 		Categorize(&cli.StringSliceFlag{
 			Name:  "copy-up",
@@ -142,7 +148,7 @@ See https://rootlesscontaine.rs/getting-started/common/ .
 		}, CategoryMount),
 		Categorize(&cli.StringFlag{
 			Name:  "port-driver",
-			Usage: "port driver for non-host network. [none, builtin, slirp4netns]",
+			Usage: "port driver for non-host network. [none, implicit (for pasta), builtin, slirp4netns]",
 			Value: "none",
 		}, CategoryPort),
 		Categorize(&cli.StringSliceFlag{
@@ -333,14 +339,14 @@ func createParentOpt(clicontext *cli.Context, pipeFDEnvKey, stateDirEnvKey, pare
 	ipv6 := clicontext.Bool("ipv6")
 	if ipv6 {
 		logrus.Warn("ipv6 is experimental")
-		if s := clicontext.String("net"); s != "slirp4netns" {
+		if s := clicontext.String("net"); s != "pasta" && s != "slirp4netns" {
 			logrus.Warnf("--ipv6 is discarded for --net=%s", s)
 		}
 	}
 
 	disableHostLoopback := clicontext.Bool("disable-host-loopback")
 	if !disableHostLoopback && clicontext.String("net") != "host" {
-		logrus.Warn("specifying --disable-host-loopback is highly recommended to prohibit connecting to 127.0.0.1:* on the host namespace (requires slirp4netns or VPNKit)")
+		logrus.Warn("specifying --disable-host-loopback is highly recommended to prohibit connecting to 127.0.0.1:* on the host namespace (requires pasta, slirp4netns, or VPNKit)")
 	}
 
 	slirp4netnsAPISocketPath := ""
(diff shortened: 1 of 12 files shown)


sbrivio-rh:
I just found this -- let me know if you need any support with it!

AkihiroSuda:
> I just found this -- let me know if you need any support with it!

Thanks, the current blocker of this PR is how to dynamically expose and unexpose ports, without restarting the pasta process.
Wondering if pasta could support an API socket as in slirp4netns, or just support dynamically reloading a configuration file with inotify, or maybe just with SIGHUP.

sbrivio-rh:
> Thanks, the current blocker of this PR is how to dynamically expose and unexpose ports, without restarting the pasta process. Wondering if pasta could support an API socket as in slirp4netns, or just support dynamically reloading a configuration file with inotify, or maybe just with SIGHUP.

`pasta` supports automatic port forwarding, from the init namespace based on ports bound in the target namespace (`-t auto` for TCP with a periodic one-second scan, `-u auto` for UDP [without periodic scan](https://bugs.passt.top/show_bug.cgi?id=45)), and from the target namespace to the init namespace, over the loopback interface if addressing allows (`-T auto` and `-U auto`).

So far, we didn't add any possibility of dynamic and explicit, user-supplied port forwarding configuration, because of these potential downsides with regard to security aspects:

* configuration file
  * requires parsing at runtime
  * needs one further file handle to be pre-opened (pasta has no access to the filesystem after the configuration phase), meaning additional rules in AppArmor policies and SELinux type enforcements

* socket API
  * requires parsing at runtime
  * needs one further pre-opened socket which requires a number of access control considerations
  * also implies additional rules in AppArmor and SELinux policies

I'm not claiming it's impossible to implement this in a secure way, I'm just saying it's much harder.

I understand this might cause some headaches in integrations, but is the automatic port forwarding really not good enough for rootlesskit use cases?

If it's not, I would rather consider a binary configuration file, just for ports, possibly four different filesystem entries (TCP and UDP over IPv4 and IPv6), reloaded via inotify (we already use one inotify listener in pasta, to detect that a filesystem-bound namespace is going away and exit).

The user would create those files, pass them as command line options, and also be in charge of its permissions. In AppArmor and SELinux rules, we would just need to make sure that pasta is able to read from those files.

Cc: @dgibson

AkihiroSuda:
Thanks for explanation

> I understand this might cause some headaches in integrations, but is the automatic port forwarding really not good enough for rootlesskit use cases?

Not enough for RootlessKit, as RootlessKit adopts an explicit port API: https://github.com/rootless-containers/rootlesskit/blob/v1.1.0/pkg/api/openapi.yaml#L20-L54

AkihiroSuda:
The default value of `fs.inotify.max_user_instances` seems relatively tight on Ubuntu (128), so reloading a config file on `SIGHUP` might be better than watching inotify.

sbrivio-rh:
> The default value of `fs.inotify.max_user_instances` seems relatively tight on Ubuntu (128), so reloading a config file on `SIGHUP` might be better than watching inotify.

Well, we might already use one inotify instance (unless `--no-netns-quit` is given), even though not in your integration as you're using `"--netns", fmt.Sprintf("/proc/%d/ns/net", childPID),`. In general terms, we could reuse the same instance for any additional files we want to monitor.

The `SIGHUP` approach comes with some additional problems:
- we would need to consider and define who can send `SIGHUP`, also in LSM policies
- pasta is currently single-threaded, and if we don't block the signal, we'll have some non-trivial amount of operations potentially interrupting, temporarily, data transfers. If we block the signal:
  - we would need to add one additional system call to the allowed set in the seccomp filter (sigtimedwait(2), which we need to fetch pending signals)
  - we would get additional overhead from periodically calling sigtimedwait()

The inotify instance is poll()able, so it integrates quite naturally with the existing event model. Unless inotify is really unusable on Ubuntu, I'd rather stick to that.

cmd/rootlesskit/main.go
```diff
@@ -108,24 +114,24 @@ See https://rootlesscontaine.rs/getting-started/common/ .
...
 		}, CategoryNetwork),
 		Categorize(&cli.StringFlag{
 			Name:  "cidr",
-			Usage: "CIDR for slirp4netns network (default: 10.0.2.0/24)",
+			Usage: "CIDR for pasta and slirp4netns networks (default: 10.0.2.0/24)",
 		}, CategoryNetwork),
 		Categorize(&cli.StringFlag{
 			Name:  "ifname",
-			Usage: "Network interface name (default: tap0 for slirp4netns and vpnkit, eth0 for lxc-user-nic)",
+			Usage: "Network interface name (default: tap0 for pasta, slirp4netns, and vpnkit; eth0 for lxc-user-nic)",
```
fahedouch, line 126: typo `Network interface name (default: tap0 for pasta, slirp4netns and vpnkit; eth0 for lxc-user-nic`

sbrivio-rh: Not [necessarily](https://en.wikipedia.org/wiki/Serial_comma#Recommendations_by_style_guides), but I couldn't find other examples in this file to check for consistency (other than [another one](https://github.com/rootless-containers/rootlesskit/pull/358/files/85b34d5037fac6f82bd8aee07c976f9f2e75e782#diff-a90a85572beea5c024965d62fc2abdce609e5dac6e7607fb344075515fc9a369R329) introduced by this commit itself).

cmd/rootlesskit/main.go
```diff
@@ -313,14 +319,14 @@ func createParentOpt(clicontext *cli.Context, pipeFDEnvKey, stateDirEnvKey, pare
...
 	if ipv6 {
 		logrus.Warn("ipv6 is experimental")
-		if s := clicontext.String("net"); s != "slirp4netns" {
+		if s := clicontext.String("net"); s != "pasta" && s != "slirp4netns" {
 			logrus.Warnf("--ipv6 is discarded for --net=%s", s)
 		}
 	}
 
 	disableHostLoopback := clicontext.Bool("disable-host-loopback")
 	if !disableHostLoopback && clicontext.String("net") != "host" {
```
fahedouch, line 328: I think there is a problem here , It should be `if disableHostLoopback && clicontext.String("net") != "host"` right ?

pkg/network/pasta/pasta.go
```diff
@@ -0,0 +1,183 @@
...
+
+	"github.com/rootless-containers/rootlesskit/pkg/api"
+	"github.com/rootless-containers/rootlesskit/pkg/common"
+	"github.com/rootless-containers/rootlesskit/pkg/network"
+	"github.com/rootless-containers/rootlesskit/pkg/network/iputils"
+	"github.com/rootless-containers/rootlesskit/pkg/network/parentutils"
+)
+
+// NewParentDriver instantiates new parent driver.
+// Requires slirp4netns v0.4.0 or later.
```
fahedouch, line 23: `pasta` ?  I am not sure if past depends on `slirp4netns`

AkihiroSuda: Will fix later

pkg/network/pasta/pasta.go
```diff
@@ -0,0 +1,183 @@
...
+
+func (d *parentDriver) MTU() int {
+	return d.mtu
+}
+
+func (d *parentDriver) ConfigureNetwork(childPID int, stateDir string) (*common.NetworkMessage, func() error, error) {
+	tap := d.ifname
+	var cleanups []func() error
+	if err := parentutils.PrepareTap(childPID, tap); err != nil {
+		return nil, common.Seq(cleanups), fmt.Errorf("setting up tap %s: %w", tap, err)
```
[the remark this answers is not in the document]
AkihiroSuda: Not necessary, as the netns will be freed when the child process exits

pkg/network/pasta/pasta.go
dgibson: `pasta` does not depend on `slirp4netns`, that's kind of the point.

AkihiroSuda:
Update (18843ccc0be4e62c02c5c037414dbc14595d5c84)

- No support for explicit port forwarding (`rootlessctl add-ports`),  as pasta doesn't support it yet.
  Use `--port-driver=implicit` to let pasta forward ports implicitly.
  The forwarded ports are not visible in `rootlessctl list-ports`.

- Child can't access the parent IP

- Tested with pasta 2023_05_09.96f8d55 on Ubuntu 23.04.
  Doesn't work with 2023_06_03.429e1a7: `Option --no-copy-routes needs --config-net`
  (This is printed despite that `--no-copy-routes` is not specified)

- Doesn't work with Ubuntu 23.04's dpkg (`passt_0.0~git20230216.4663ccc-1_amd64.deb`):
  `Couldn't open user namespace /proc/51813/ns/user: Permission denied`
  Likely to be related to AppArmor.

sbrivio-rh:
> Update ([18843cc](https://github.com/rootless-containers/rootlesskit/commit/18843ccc0be4e62c02c5c037414dbc14595d5c84))

Sorry for the delay on the port forwarding functionality, and thanks for the error reports! I'm currently traveling, I'll look into them next week.

AkihiroSuda:
[2023_06_25.32660ce](https://passt.top/passt/tag/?h=2023_06_25.32660ce) seems to have resolved several issues (Thanks @sbrivio-rh)

- - -
TODO: analyze the dpkg issue

sbrivio-rh:
> [2023_06_25.32660ce](https://passt.top/passt/tag/?h=2023_06_25.32660ce) seems to have resolved several issues (Thanks @sbrivio-rh)

Oh, you already noticed :) I wanted to finish building new versions of the packages I maintain (Debian and Fedora) versions before updating this ticket...

> TODO: analyze the dpkg issue

I updated the AppArmor profiles upstream and in the Debian packages (which are then synchronised to Ubuntu) after that package version, but I don't think I changed/fixed anything related to that. I'll try to have a look.

sbrivio-rh:
@AkihiroSuda do you have some kind of timeline for version 1.1.1? I'm working on the port forwarding configuration stuff right now and I'm trying to understand if it makes sense that I try to rush a bit, a lot, or not at all :)

AkihiroSuda:
> @AkihiroSuda do you have some kind of timeline for version 2.0.0? I'm working on the port forwarding configuration stuff right now and I'm trying to understand if it makes sense that I try to rush a bit, a lot, or not at all :)

Thanks. 
No timeline, but probably it will be within 3-4 weeks.

PR #362, opened 2023-04-07
support detach-ns

fahedouch:


AkihiroSuda:
Thanks, carried as:
- https://github.com/rootless-containers/rootlesskit/pull/379

PR #369, opened 2023-05-28
Remove duplicate id ranges returned by getsubu/gid for username/uid

i-do-cpp:
In LDAP-based environments both `getsubids myusername` and `getsubids 1234` (where `id myusername` equals to `1234`) can return exactly the same subuid ranges. This happens when one of `/etc/subuid` or `nss` returns mappings for both the user id and the user name. Both `newgidmap` and `newuidmap` fail when the parameters contain duplicate ranges.

This PR fixes the problem by removing duplicates from the list of subid mappings in `dynidtools`.
`idtools` (`/etc/subuid` handling) is left untouched.

### DIFF
--- a/pkg/parent/dynidtools/dynidtools.go
+++ b/pkg/parent/dynidtools/dynidtools.go
@@ -14,6 +14,18 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func withoutDuplicates(sliceList []idtools.SubIDRange) []idtools.SubIDRange {
+	seenKeys := make(map[idtools.SubIDRange]bool)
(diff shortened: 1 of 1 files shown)


pkg/parent/dynidtools/dynidtools.go
```diff
@@ -14,6 +14,18 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func withoutDuplicates[T idtools.SubIDRange](sliceList []T) []T {
```
AkihiroSuda, line 17: Do we need `T`?

AkihiroSuda (approved):
Thanks

PR #427, opened 2024-03-24
feat: add new bridge network

antrusd:
This PR will add net `bridge` driver support. The driver almost similar with `lxc-user-nic` but without attaching connectivity between host namespace with target namespace. The connectivity need to be configured manually with the help `veth` interface (need root privilege) from host namespace which can be "watched" using `bash` script +  `systemd` service.

eg.
```bash
#!/usr/bin/env bash

export ROOTKIT_STATE_DIRECTORY=/path/to/rootkit/state/directory
export BRIDGE=docker0
export VETHP=docker0p1
export VETHC=docker0c1

# add masquerade support
echo 1 > /proc/sys/net/ipv4/ip_forward
iptables-save | grep -q ${BRIDGE} || iptables -t nat -A POSTROUTING -s 172.17.0.0/16 ! -o ${BRIDGE} -j MASQUERADE

while :; do
  until test -r ${ROOTKIT_STATE_DIRECTORY}/child_pid; do
    sleep 1
  done

  child_pid=$(cat ${ROOTKIT_STATE_DIRECTORY}/child_pid)
  mkdir -p /var/run/netns
  ln -sf /proc/${child_pid}/ns/net /var/run/netns/${BRIDGE}
  ip link add ${VETHC} type veth peer name ${VETHP}
  ip link set ${VETHP} up
  ip link set ${VETHP} master ${BRIDGE}
  ip link set ${VETHC} netns ${BRIDGE}

  while test -r ${ROOTKIT_STATE_DIRECTORY}/child_pid && ip link show ${VETHP} &>/dev/null; do
    sleep 1
  done
done
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,9 @@ all: $(addprefix bin/, $(BINARIES))
 clean:
 	$(RM) -r bin/ _artifact/
 
+bin/rootlesskit-debug: $(GO_FILES)
+	$(GO) build -o $@ -gcflags="all=-N -l" -v ./cmd/rootlesskit
+
 bin/rootlesskit: $(GO_FILES)
 	$(GO) build -o $@ -v ./cmd/rootlesskit
 
--- a/cmd/rootlesskit/main.go
+++ b/cmd/rootlesskit/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/rootless-containers/rootlesskit/v2/pkg/network/pasta"
 	"github.com/rootless-containers/rootlesskit/v2/pkg/network/slirp4netns"
 	"github.com/rootless-containers/rootlesskit/v2/pkg/network/vpnkit"
+	"github.com/rootless-containers/rootlesskit/v2/pkg/network/bridge"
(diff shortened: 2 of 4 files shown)


AkihiroSuda:
I'd rather prefer to have `rootlesskit --net=none` that would just leaves netns empty, as the connectivity has to be configured by an external script

Makefile
```diff
@@ -9,6 +9,9 @@ all: $(addprefix bin/, $(BINARIES))
 clean:
 	$(RM) -r bin/ _artifact/
 
+bin/rootlesskit-debug: $(GO_FILES)
```
AkihiroSuda, line 12: This should be a separate PR, and I'd rather prefer to use a variable like `GO_BUILD_FLAGS`

antrusd:
Actually `none`  was the option I looking for in the first time, but forgot why I end up with `bridge` proposal. Will close this and will propose `none` network here https://github.com/rootless-containers/rootlesskit/pull/430 .

Makefile
antrusd: agree

Issue #428, opened 2024-03-28
Sockets from systemd activation not inherited by child


charliemirabile:
In order to make it easier to run daemons that normally run as root as a non root user with rootlesskit, it would be very helpful for rootless kit to be aware of and support the systemd socket activation (https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html).

The gist is that socket(s) are created by systemd instead of the daemon itself and when a process attempts to connect to the socket(s), the actual daemon is started and it inherits the socket(s) and special environment variables are set to describe who the socket(s) are intended for, how many file descriptors there are, and their names. The daemon (if it is aware of this feature and properly configured to work with it) can then skip setting sockets as it normally might during startup and instead use the provided one(s).

The benefits are myriad: using socket activation can cut down on startup time because daemons are only started when the first client attempts to connect to them, and further because the daemon inherits already configured sockets, all of its code can run unprivileged because it can rely on systemd can use its permissions to bind to low port numbers.

In terms of what is required for this feature to work in rootlesskit, the parent process would need to check for the presence of the systemd environment variables (`$LISTEN_FDS`, `$LISTEN_PID`, `$LISTEN_FDNAMES`) and if it detects them, make sure that the file descriptors 3, 4, ... (3 + $LISTEN_FDS -1) are inherited by the child, and then the child would again need to make sure that those file descriptors are inherited by the real program, and also fixup the value of LISTEN_PID to reflect the pid of the child process before actually execve'ing it.

charliemirabile:
It seems like what I suggested about fixing up the pid in LISTEN_PID is actually quite annoying to do in go because fork and exec are tightly coupled. I think this is ok though because if the daemon in question is picky about the value of LISTEN_PID (like the docker-daemon) a script can be inserted that does the following to fixup the value for the daemon
```sh
#!/bin/sh
export LISTEN_PID=$$
exec <real daemon> $@
```

PR #429, opened 2024-03-28
Support systemd socket activation

charliemirabile:
Fixes: #428

This is my first time writing go code, so go easy on me :)

I tried to break this down into a clear set of commits, so the actual logic I want to implement is easy to follow separate from the necessary refactoring I did. Happy to squash this down into one commit though if that is the etiquette around here.

### DIFF
--- a/hack/integration-systemd-socket.sh
+++ b/hack/integration-systemd-socket.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+set -e
+if [ -z "$EXECED" ]
+then
+	systemd-socket-activate -E EXECED=1 -l /tmp/activate.sock socat ACCEPT-FD:3 EXEC:"rootlesskit $0",nofork 2>/dev/null &
+	OUTPUT="$(curl --unix-socket /tmp/activate.sock http://localhost/hello 2>/dev/null)"
+	[ "$(printf 'Hello\n' )" = "$OUTPUT" ] || exit 1
+else
+	[ "$LISTEN_FDS" = "1" ] || exit 1
+	read -r REQUEST
+	if [ "$(printf 'GET /hello HTTP/1.1\r\n')" = "$REQUEST" ]
+	then
+	    printf 'HTTP/1.1 200 OK\r\nContent-Length: 6\r\n\r\nHello\n'
+	else
+	    printf 'HTTP/1.1 400 Bad Request\r\nContent-Length: 5\r\n\r\nBad!\n'
+	fi
+fi
--- a/pkg/child/child.go
+++ b/pkg/child/child.go
@@ -34,6 +34,23 @@ var propagationStates = map[string]uintptr{
 	"rslave":   uintptr(unix.MS_REC | unix.MS_SLAVE),
 }
 
+func setupFiles(cmd *exec.Cmd) {
+	// 0 1 and 2  are used for stdin. stdout, and stderr
+	const firstExtraFD = 3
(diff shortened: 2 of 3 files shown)


AkihiroSuda:
Thanks, how to test this?

AkihiroSuda:
Thanks, is it testable without depending on Docker?
Can we have a test script like https://github.com/rootless-containers/rootlesskit/blob/v2.0.2/hack/integration-evacuate-cgroup2.sh#L4 ?

charliemirabile:
yes, you could write a script like this:
```sh
#!/bin/sh
read -r REQUEST
printf 'GET /hello HTTP/1.1\r\n' | xxd >&2
echo "$REQUEST" | xxd >&2
if [ "$(printf 'GET /hello HTTP/1.1\r\n')" = "$REQUEST" ]
then
    printf 'HTTP/1.1 200 OK\r\nContent-Length: 6\r\n\r\nHello\n'
else
    printf 'HTTP/1.1 400 Bad Request\r\nContent-Length: 5\r\n\r\nBad!\n'
fi
```
and then do `systemd-socket-activate -l /tmp/test.sock socat ACCEPT-FD:3 EXEC:./test.sh,nofork` and then verify that you get `Hello` back from `curl --unix-socket /tmp/test.sock https://localhost/hello`

charliemirabile:
Ok, I have a working testing script, but I am struggling to add it to the existing integration testing framework. 
```diff
diff --git a/hack/integration-systemd-socket.sh b/hack/integration-systemd-socket.sh
new file mode 100755
index 0000000..1129a72
--- /dev/null
+++ b/hack/integration-systemd-socket.sh
@@ -0,0 +1,17 @@
+#!/bin/sh
+set -e
+if [ -z "$EXECED" ]
+then
+       systemd-socket-activate -E EXECED=1 -l /tmp/activate.sock socat ACCEPT-FD:3 EXEC:"rootlesskit $0",nofork 2>/dev/null &
+       OUTPUT="$(curl --unix-socket /tmp/activate.sock http://localhost/hello 2>/dev/null)"
+       [ "$(printf 'Hello\n' )" = "$OUTPUT" ] || exit 1
+else
+       [ "$LISTEN_FDS" = "1" ] || exit 1
+       read -r REQUEST
+       if [ "$(printf 'GET /hello HTTP/1.1\r\n')" = "$REQUEST" ]
+       then
+           printf 'HTTP/1.1 200 OK\r\nContent-Length: 6\r\n\r\nHello\n'
+       else
+           printf 'HTTP/1.1 400 Bad Request\r\nContent-Length: 5\r\n\r\nBad!\n'
+       fi
+fi
```
I need a newer version of socat (v1.8.0 instead of v.1.7.4) which is available in ubuntu 24.04 but not 22.04. I feel like bumping the ubuntu version used in the Dockerfile across the board would probably be outside of the scope of this PR (especially because it isn't trivial to do). Is it ok for me to just add the testing script without adding it to the CI, or should I make a separate PR and try to bump the ubuntu version and then come back to this?

AkihiroSuda:
Integration to CI can be worked out later.
https://github.com/rootless-containers/rootlesskit/blob/v2.0.2/hack/integration-evacuate-cgroup2.sh isn’t integrated to CI either, as the current Dockerfile lacks systemd

charliemirabile:
Alright, just pushed one more commit with the script

AkihiroSuda (approved):
Thanks

PR #430, opened 2024-04-02
feat: add new none network

antrusd:
This PR will add net `none` driver support. The driver almost similar with `host` but only provide `lo` interface configured. This is follow up from https://github.com/rootless-containers/rootlesskit/pull/427 .

### DIFF
--- a/cmd/rootlesskit/main.go
+++ b/cmd/rootlesskit/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/rootless-containers/rootlesskit/v2/pkg/network/pasta"
 	"github.com/rootless-containers/rootlesskit/v2/pkg/network/slirp4netns"
 	"github.com/rootless-containers/rootlesskit/v2/pkg/network/vpnkit"
+	"github.com/rootless-containers/rootlesskit/v2/pkg/network/none"
 	"github.com/rootless-containers/rootlesskit/v2/pkg/parent"
 	"github.com/rootless-containers/rootlesskit/v2/pkg/port/builtin"
 	"github.com/rootless-containers/rootlesskit/v2/pkg/port/portutil"
@@ -83,7 +84,7 @@ See https://rootlesscontaine.rs/getting-started/common/ .
 		}, CategoryState),
 		Categorize(&cli.StringFlag{
 			Name:  "net",
-			Usage: "network driver [host, pasta(experimental), slirp4netns, vpnkit, lxc-user-nic(experimental)]",
+			Usage: "network driver [host, none, pasta(experimental), slirp4netns, vpnkit, lxc-user-nic(experimental)]",
 			Value: "host",
 		}, CategoryNetwork),
 		Categorize(&cli.StringFlag{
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
```console
$ rootlesskit --net=none bash
WARN[0000] [rootlesskit:parent] specifying --disable-host-loopback is highly recommended to prohibit connecting to 127.0.0.1:* on the host namespace (requires pasta, slirp4netns, or VPNKit) 
```

This warning should not be printed for `--net-none`

cmd/rootlesskit/main.go
```diff
@@ -388,6 +389,26 @@ func createParentOpt(clicontext *cli.Context, pipeFDEnvKey, stateDirEnvKey, pare
...
+			return opt, errors.New("custom cidr is not supported for --net=none")
+		}
+		if ifname != "" {
+			return opt, errors.New("ifname cannot be specified for --net=none")
+		}
+		switch portDriver := clicontext.String("port-driver"); portDriver {
+		case "none":
+			// NOP
+		default:
+			return opt, errors.New("network \"none\" requires port driver \"none\"")
```
AkihiroSuda, line 406: "builtin" may work too?

pkg/network/none/none.go
```diff
@@ -0,0 +1,60 @@
...
+
+	cmds := [][]string{
+		[]string{"nsenter", "-t", strconv.Itoa(childPID), "--no-fork", "-n", "-m", "-U", "--preserve-credentials", "ip", "address", "add", "127.0.0.1/8", "dev", "lo"},
+		[]string{"nsenter", "-t", strconv.Itoa(childPID), "--no-fork", "-n", "-m", "-U", "--preserve-credentials", "ip", "link", "set", "lo", "up"},
+	}
+	if err := common.Execs(os.Stderr, os.Environ(), cmds); err != nil {
+		return nil, nil, err
+	}
+
+	if detachedNetNSPath != "" {
```
AkihiroSuda, line 46: Got an error:
```console
$ rootlesskit --net=none --detach-netns bash
WARN[0000] [rootlesskit:parent] specifying --disable-host-loopback is highly recommended to prohibit connecting to 127.0.0.1:* on the host namespace (requires pasta, slirp4netns, or VPNKit) 
nsenter: reassociate to namespace 'ns/net' failed: Operation not permitted
[rootlesskit:parent] error: failed to setup network &{}: exit status 1
[rootlesskit:child ] error: EOF
```

antrusd:
> ```
> $ rootlesskit --net=none bash
> WARN[0000] [rootlesskit:parent] specifying --disable-host-loopback is highly recommended to prohibit connecting to 127.0.0.1:* on the host namespace (requires pasta, slirp4netns, or VPNKit) 
> ```
> 
> This warning should not be printed for `--net-none`

Updated [here](https://github.com/rootless-containers/rootlesskit/pull/430/files#diff-a90a85572beea5c024965d62fc2abdce609e5dac6e7607fb344075515fc9a369R372)

cmd/rootlesskit/main.go
antrusd: right, updated

pkg/network/none/none.go
antrusd: hmm, I never use `--detach-netns` actually, missed the test.

but then `none` network driver shouldn't have `--detach-netns` support since we don't have any network but `loopback` ?

AkihiroSuda: `none` may support `--detach-netns` for consistency with other network drivers, but not a hard requirement.

antrusd: if that so, then configuring `loppback` when `--detach-netns` seems useless.

AkihiroSuda: `rootlesskit --net=none --detach-netns bash` will run `bash` in the host netns, and detach a new lo-only netns in `$ROOTLESSKIT_STATE_DIR/netns`.

antrusd: got it, updated

AkihiroSuda (approved):
Thanks

PR #458, opened 2024-08-16
Let pasta configure interface, fix IPv6 outbound connectivity

sbrivio-rh:
Fixes https://github.com/moby/moby/issues/48257

### DIFF
--- a/pkg/child/child.go
+++ b/pkg/child/child.go
@@ -215,8 +215,11 @@ func setupNet(stateDir string, msg *messages.ParentInitNetworkDriverCompleted, e
 		if err := os.WriteFile(stateDirResolvConf, generateResolvConf(msg.DNS), 0644); err != nil {
 			return fmt.Errorf("writing %s: %w", stateDirResolvConf, err)
 		}
-		if err := activateDev(dev, msg.IP, msg.Netmask, msg.Gateway, msg.MTU); err != nil {
-			return err
+		Info, _ := driver.ChildDriverInfo()
+		if !Info.ConfiguresInterface {
+			if err := activateDev(dev, msg.IP, msg.Netmask, msg.Gateway, msg.MTU); err != nil {
+				return err
+			}
 		}
 		if etcWasCopied {
 			// remove copied-up link
@@ -255,7 +258,11 @@ func setupNet(stateDir string, msg *messages.ParentInitNetworkDriverCompleted, e
 			return fmt.Errorf("writing %s: %w", stateDirResolvConf, err)
 		}
 		if err := ns.WithNetNSPath(detachedNetNSPath, func(_ ns.NetNS) error {
-			return activateDev(dev, msg.IP, msg.Netmask, msg.Gateway, msg.MTU)
+			Info, _ := driver.ChildDriverInfo()
+			if !Info.ConfiguresInterface {
+				return activateDev(dev, msg.IP, msg.Netmask, msg.Gateway, msg.MTU)
+			}
+			return nil
 		}); err != nil {
 			return err
 		}
--- a/pkg/network/network.go
+++ b/pkg/network/network.go
@@ -17,6 +17,10 @@ type ParentDriver interface {
 	ConfigureNetwork(childPID int, stateDir, detachedNetNSPath string) (netmsg *messages.ParentInitNetworkDriverCompleted, cleanup func() error, err error)
 }
 
+type ChildDriverInfo struct {
+	ConfiguresInterface bool // Driver configures own namespace interface
+}
+
 // ChildDriver is called from the child namespace
 type ChildDriver interface {
 	// ConfigureNetworkChild is executed in the child's namespaces, excluding detached-netns.
@@ -24,4 +28,6 @@ type ChildDriver interface {
 	// netmsg MAY be modified.
 	// devName is like "tap" or "eth0"
 	ConfigureNetworkChild(netmsg *messages.ParentInitNetworkDriverCompleted, detachedNetNSPath string) (devName string, err error)
+
+	ChildDriverInfo() (*ChildDriverInfo, error)
 }
--- a/pkg/network/pasta/pasta.go
+++ b/pkg/network/pasta/pasta.go
@@ -111,7 +111,6 @@ func (d *parentDriver) ConfigureNetwork(childPID int, stateDir, detachedNetNSPat
 	}
 
 	opts := []string{
-		"--foreground",
 		"--stderr",
 		"--ns-ifname=" + d.ifname,
 		"--mtu=" + strconv.Itoa(d.mtu),
@@ -147,21 +146,18 @@ func (d *parentDriver) ConfigureNetwork(childPID int, stateDir, detachedNetNSPat
 	// `Couldn't open user namespace /proc/51813/ns/user: Permission denied`
 	// Possibly related to AppArmor.
 	cmd := exec.Command(d.binary, opts...)
-	cmd.Stdout = d.logWriter
-	cmd.Stderr = d.logWriter
-	cleanups = append(cleanups, func() error {
-		logrus.Debugf("killing pasta")
-		if cmd.Process != nil {
-			_ = cmd.Process.Kill()
-		}
-		wErr := cmd.Wait()
-		logrus.Debugf("killed pasta: %v", wErr)
-		return nil
-	})
 	logrus.Debugf("Executing %v", cmd.Args)
-	if err := cmd.Start(); err != nil {
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		exitErr := &exec.ExitError{}
+		if errors.As(err, &exitErr) {
+			return nil, common.Seq(cleanups),
+			       fmt.Errorf("pasta failed with exit code %d:\n%s",
+					  exitErr.ExitCode(), string(out))
+		}
 		return nil, common.Seq(cleanups), fmt.Errorf("executing %v: %w", cmd, err)
 	}
+
 	netmsg := messages.ParentInitNetworkDriverCompleted{
 		Dev: tap,
 		MTU: d.mtu,
--- a/pkg/network/pasta/pasta.go
+++ b/pkg/network/pasta/pasta.go
@@ -17,7 +17,6 @@ import (
 	"github.com/rootless-containers/rootlesskit/v2/pkg/messages"
 	"github.com/rootless-containers/rootlesskit/v2/pkg/network"
 	"github.com/rootless-containers/rootlesskit/v2/pkg/network/iputils"
-	"github.com/rootless-containers/rootlesskit/v2/pkg/network/parentutils"
 )
 
 // NewParentDriver instantiates new parent driver.
@@ -92,9 +91,6 @@ func (d *parentDriver) MTU() int {
 func (d *parentDriver) ConfigureNetwork(childPID int, stateDir, detachedNetNSPath string) (*messages.ParentInitNetworkDriverCompleted, func() error, error) {
 	tap := d.ifname
 	var cleanups []func() error
-	if err := parentutils.PrepareTap(childPID, detachedNetNSPath, tap); err != nil {
-		return nil, common.Seq(cleanups), fmt.Errorf("setting up tap %s: %w", tap, err)
-	}
 
 	address, err := iputils.AddIPInt(d.ipnet.IP, 100)
 	if err != nil {
@@ -114,8 +110,7 @@ func (d *parentDriver) ConfigureNetwork(childPID int, stateDir, detachedNetNSPat
 		"--stderr",
 		"--ns-ifname=" + d.ifname,
 		"--mtu=" + strconv.Itoa(d.mtu),
-		"--no-dhcp",
-		"--no-ra",
+		"--config-net",
 		"--address=" + address.String(),
 		"--netmask=" + strconv.Itoa(netmask),
 		"--gateway=" + gateway.String(),
@@ -185,6 +180,12 @@ func NewChildDriver() network.ChildDriver {
 }
 
 type childDriver struct {
+}
+
+func (d *childDriver) ChildDriverInfo() (*network.ChildDriverInfo, error) {
+	return &network.ChildDriverInfo {
+		ConfiguresInterface: true,
+	}, nil
 }
 
 func (d *childDriver) ConfigureNetworkChild(netmsg *messages.ParentInitNetworkDriverCompleted, detachedNetNSPath string) (string, error) {
--- a/pkg/network/lxcusernic/lxcusernic.go
+++ b/pkg/network/lxcusernic/lxcusernic.go
@@ -148,6 +148,12 @@ func exchangeDHCP(c *client4.Client, dev string, detachedNetNSPath string) (*dhc
 	return ack, nil
 }
 
+func (d *childDriver) ChildDriverInfo() (*network.ChildDriverInfo, error) {
+	return &network.ChildDriverInfo {
+		ConfiguresInterface: false,
+	}, nil
+}
+
 func (d *childDriver) ConfigureNetworkChild(netmsg *messages.ParentInitNetworkDriverCompleted, detachedNetNSPath string) (string, error) {
 	dev := netmsg.Dev
 	if dev == "" {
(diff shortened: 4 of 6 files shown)


pkg/child/child.go
```diff
@@ -187,7 +187,7 @@ func setupCopyDir(driver copyup.ChildDriver, dirs []string) (bool, error) {
 // setupNet sets up the network driver.
 //
 // NOTE: msg is altered during calling driver.ConfigureNetworkChild
-func setupNet(stateDir string, msg *messages.ParentInitNetworkDriverCompleted, etcWasCopied bool, driver network.ChildDriver, detachedNetNSPath string) error {
+func setupNet(stateDir string, msg *messages.ParentInitNetworkDriverCompleted, etcWasCopied bool, driver network.ChildDriver, driverConfiguresNet bool, detachedNetNSPath string) error {
```
AkihiroSuda, line 190: Probably `ChildDriver` should have a method something like `ChildDriverInfo()` , then `setupNet` does not need to have the `driverConfiguresNet` arg

sbrivio-rh:
Oops, looking into UDP failure now...

pkg/network/pasta/pasta.go
```diff
@@ -143,9 +137,14 @@ func (d *parentDriver) ConfigureNetwork(childPID int, stateDir, detachedNetNSPat
...
-	// `Couldn't open user namespace /proc/51813/ns/user: Permission denied`
-	// Possibly related to AppArmor.
+	// FIXME: Doesn't work with:
+	// - passt-0.0~git20230627.289301b-1 (Ubuntu 23.10)
+	// - passt-0.0~git20240220.1e6f92b-1 (Ubuntu 24.04)
+	// see https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2077158
+	//
+	// Workaround: set the kernel.apparmor_restrict_unprivileged_userns
+	// sysctl to 0, or (preferred) add the AppArmor profile from upstream,
+	// or from Debian packages, or from Ubuntu > 24.10.
```
AkihiroSuda, line 147: Could you also update https://github.com/rootless-containers/rootlesskit/blob/master/docs/network.md#--netpasta-experimental

AkihiroSuda:
CI is failing https://github.com/rootless-containers/rootlesskit/actions/runs/10418395143/job/28854469268?pr=458
```
+ rootlesskit --net=pasta --copy-up=/etc --copy-up=/run --disable-host-loopback -- nslookup example.com
time="2024-0[8](https://github.com/rootless-containers/rootlesskit/actions/runs/10418395143/job/28854469268?pr=458#step:15:9)-16T10:22:22Z" level=warning msg="[rootlesskit:parent] \"pasta\" network driver is experimental. Needs very recent version of pasta (see docs/network.md)."
;; UDP setup with 10.0.2.3#53(10.0.2.3) for example.com failed: network unreachable.
;; no servers could be reached

;; UDP setup with [10](https://github.com/rootless-containers/rootlesskit/actions/runs/10418395143/job/28854469268?pr=458#step:15:11).0.2.3#53(10.0.2.3) for example.com failed: network unreachable.
;; no servers could be reached

;; UDP setup with 10.0.2.3#53(10.0.2.3) for example.com failed: network unreachable.
;; no servers could be reached


[rootlesskit:child ] error: command [nslookup example.com] exited: exit status 1
[rootlesskit:parent] error: child exited: exit status 1
```

pkg/network/pasta/pasta.go
```diff
@@ -112,11 +108,9 @@ func (d *parentDriver) ConfigureNetwork(childPID int, stateDir, detachedNetNSPat
 
 	opts := []string{
 		"--foreground",
-		"--stderr",
```
AkihiroSuda, line 115: ?

sbrivio-rh:
> Oops, looking into UDP failure now...

Weird, that works for me on Debian unstable, passt-0.0~git20240814.61c0b0d-1:

```
$ rootlesskit --net=pasta --copy-up=/etc --copy-up=/run --disable-host-loopback -- nslookup example.com
WARN[0000] [rootlesskit:parent] "pasta" network driver is experimental. Needs very recent version of pasta (see docs/network.md). 
Server:		10.0.2.3
Address:	10.0.2.3#53

Non-authoritative answer:
Name:	example.com
Address: 93.184.215.14
Name:	example.com
Address: 2606:2800:21f:cb07:6820:80da:af6b:8b2c
```

pkg/network/pasta/pasta.go
AkihiroSuda: The AppArmor stuff can be also a separate PR

sbrivio-rh:
> CI is failing https://github.com/rootless-containers/rootlesskit/actions/runs/10418395143/job/28854469268?pr=458

I can reproduce this using the `rootlesskit:test-integration` image. The outcome depends on the timing, sometimes I get:
```
$ rootlesskit --net=pasta --copy-up=/etc --copy-up=/run --disable-host-loopback -- nslookup example.com
WARN[0000] [rootlesskit:parent] "pasta" network driver is experimental. Needs very recent version of pasta (see docs/network.md). 
;; UDP setup with 10.0.2.3#53(10.0.2.3) for example.com failed: network unreachable.
;; no servers could be reached

;; UDP setup with 10.0.2.3#53(10.0.2.3) for example.com failed: network unreachable.
;; no servers could be reached

;; UDP setup with 10.0.2.3#53(10.0.2.3) for example.com failed: network unreachable.
;; no servers could be reached
```

sometimes:

```
$ rootlesskit --net=pasta --copy-up=/etc --copy-up=/run --disable-host-loopback -- nslookup example.com
WARN[0000] [rootlesskit:parent] "pasta" network driver is experimental. Needs very recent version of pasta (see docs/network.md). 
;; UDP setup with 10.0.2.3#53(10.0.2.3) for example.com failed: network unreachable.
;; no servers could be reached

;; UDP setup with 10.0.2.3#53(10.0.2.3) for example.com failed: network unreachable.
;; no servers could be reached

Server:		10.0.2.3
Address:	10.0.2.3#53

Non-authoritative answer:
Name:	example.com
Address: 93.184.215.14
Name:	example.com
Address: 2606:2800:21f:cb07:6820:80da:af6b:8b2c
```

(second attempt worked here). And sometimes:

```
$ rootlesskit --net=pasta --copy-up=/etc --copy-up=/run --disable-host-loopback -- nslookup example.com
WARN[0000] [rootlesskit:parent] "pasta" network driver is experimental. Needs very recent version of pasta (see docs/network.md). 
Server:		10.0.2.3
Address:	10.0.2.3#53

Non-authoritative answer:
Name:	example.com
Address: 93.184.215.14
Name:	example.com
Address: 2606:2800:21f:cb07:6820:80da:af6b:8b2c
```

that's because if we drop the invocation of a number of `ip(8)` commands, the container will be ready sooner, but pasta isn't necessarily ready to handle traffic.

The problem here is that with `--foreground` and `cmd.Start()`, you don't have a synchronisation point: pasta forks when it's ready. But if it doesn't fork, you don't know when.

I would switch this to the same [mechanism](https://github.com/containers/common/blob/99366677828a45b2ff465aa26c2c8fdb6ef8ad46/libnetwork/pasta/pasta_linux.go#L70) that Podman uses, that is, `exec.Command()` without `--foreground`. Once we're done with `exec.Command()`, we know we can proceed, and pasta will exit anyway once the namespace is gone, so there's no need for explicit cleanup. Would you see any issue with that?

AkihiroSuda:
> I would switch this to the same [mechanism](https://github.com/containers/common/blob/99366677828a45b2ff465aa26c2c8fdb6ef8ad46/libnetwork/pasta/pasta_linux.go#L70) that Podman uses, that is, exec.Command() without --foreground. Once we're done with exec.Command(), we know we can proceed, and pasta will exit anyway once the namespace is gone, so there's no need for explicit cleanup. Would you see any issue with that?

SGTM, if an abnormal exit status of `pasta` can be still caught

sbrivio-rh:
> > I would switch this to the same [mechanism](https://github.com/containers/common/blob/99366677828a45b2ff465aa26c2c8fdb6ef8ad46/libnetwork/pasta/pasta_linux.go#L70) that Podman uses, that is, exec.Command() without --foreground. Once we're done with exec.Command(), we know we can proceed, and pasta will exit anyway once the namespace is gone, so there's no need for explicit cleanup. Would you see any issue with that?
> 
> SGTM, if an abnormal exit status of `pasta` can be still caught

Yes, that's done anyway by CombinedOutput(). I just pushed the relevant commit. I still have this open:

> Probably `ChildDriver` should have a method like `ChildDriverInfo()` , then `setupNet` does not need to have the `driverConfiguresNet` arg

but the rest should be fixed.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG GO_VERSION=1.22
+ARG GO_VERSION=1.23
```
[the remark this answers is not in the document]
sbrivio-rh: Yes, I just happened to go back a bit with a rebase and SHA changed, fixed now.

sbrivio-rh:
> I still have this open:
> 
> > Probably `ChildDriver` should have a method like `ChildDriverInfo()` , then `setupNet` does not need to have the `driverConfiguresNet` arg

Added as WIP commit, I'm not quite sure what I'm doing (I use Go very rarely) but it actually works, polishing up now.

pkg/api/api.go
```diff
@@ -32,3 +33,8 @@ type PortDriverInfo struct {
 	Protos                  []string `json:"protos"`
 	DisallowLoopbackChildIP bool     `json:"disallowLoopbackChildIP,omitempty"` // since API v1.1.1
 }
+
+type ChildDriverInfo struct {
+	ConfiguresInterface	bool     `json:"configuresInterface"`
+}
```
AkihiroSuda, line 39: I don't think that this has to be published to the REST API

sbrivio-rh: Right, I also wanted to avoid that, but I have no idea where else to define this in a way that can be used by all the drivers.

pkg/network/pasta/pasta.go
```diff
@@ -147,21 +141,18 @@ func (d *parentDriver) ConfigureNetwork(childPID int, stateDir, detachedNetNSPat
...
+	if err != nil {
+		exitErr := &exec.ExitError{}
+		if errors.As(err, &exitErr) {
+			return nil, common.Seq(cleanups),
+			       fmt.Errorf("pasta failed with exit code %d:\n%s",
+					  exitErr.ExitCode(), string(out))
+		}
 		return nil, common.Seq(cleanups), fmt.Errorf("executing %v: %w", cmd, err)
 	}
+
```
AkihiroSuda, line 155: What will happen if the background pasta process dies after several minutes/hours/days?
Can RootlessKit detect that death?

pkg/api/api.go
```diff
@@ -32,3 +33,7 @@ type PortDriverInfo struct {
 	Protos                  []string `json:"protos"`
 	DisallowLoopbackChildIP bool     `json:"disallowLoopbackChildIP,omitempty"` // since API v1.1.1
 }
+
+type ChildDriverInfo struct {
+	ConfiguresInterface	bool     `json:"configuresInterface"`
```
AkihiroSuda, line 38: Needs a comment line to explain what this means

pkg/api/api.go
```diff
@@ -32,3 +33,7 @@ type PortDriverInfo struct {
 	Protos                  []string `json:"protos"`
 	DisallowLoopbackChildIP bool     `json:"disallowLoopbackChildIP,omitempty"` // since API v1.1.1
 }
+
+type ChildDriverInfo struct {
```
AkihiroSuda, line 37: This shouldn't be part of the REST API and should be moved to `pkg/network`

pkg/network/pasta/pasta.go
sbrivio-rh: It can't, but if it happens (even though we haven't received this kind of "stability" bug reports for quite a long time now), we would like users to report an issue, because anyway context will be lost (TCP connections, at least), rather than hide this by restarting the process or similar.

AkihiroSuda: The pasta process can be killed by the kernel mostly due to OOM, and in that case RootlessKit should probably abort, or at least print an error message.

pkg/network/network.go
```diff
@@ -24,4 +24,6 @@ type ChildDriver interface {
 	// netmsg MAY be modified.
 	// devName is like "tap" or "eth0"
 	ConfigureNetworkChild(netmsg *messages.ParentInitNetworkDriverCompleted, detachedNetNSPath string) (devName string, err error)
+
+	ChildDriverInfo() (*api.ChildDriverInfo, error)
```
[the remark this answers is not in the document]
sbrivio-rh: Done.

pkg/api/api.go
sbrivio-rh: Added (in network.go, now).

pkg/api/api.go
sbrivio-rh: Thanks for the tip, moved there.

pkg/network/pasta/pasta.go
```diff
@@ -147,21 +141,18 @@ func (d *parentDriver) ConfigureNetwork(childPID int, stateDir, detachedNetNSPat
 	// `Couldn't open user namespace /proc/51813/ns/user: Permission denied`
 	// Possibly related to AppArmor.
 	cmd := exec.Command(d.binary, opts...)
-	cmd.Stdout = d.logWriter
-	cmd.Stderr = d.logWriter
```
AkihiroSuda, line 151: Where is the stderr now ?

pkg/network/pasta/pasta.go
sbrivio-rh: It's pretty much impossible because it doesn't allocate memory dynamically, so I don't think it can actually be picked as a candidate by the OOM killer.

Anyway, the network interface will go away in that case, because we close the tap file descriptor, so something like that could actually be added.

I don't think it's actually needed, because the network interface going away is very obvious, but I guess I can implement it if needed (that will take a bit longer though).

pkg/network/pasta/pasta.go
sbrivio-rh: It's all in `out` -- that's what `out, err := cmd.CombinedOutput()` does.

sbrivio-rh: https://pkg.go.dev/os/exec#Cmd.CombinedOutput

AkihiroSuda (approved):
Thanks

pkg/network/pasta/pasta.go
sbrivio-rh: @AkihiroSuda I'm trying to implement this with the [`LinkSubscribe()`](https://pkg.go.dev/github.com/vishvananda/netlink#LinkSubscribe) method from github.com/vishvananda/netlink.

That part looks trivial, but I don't know where to add this kind of monitor in rootlesskit. I thought there was something like that for slirp4netns' `--exit-fd`, but I can't find it. How would you detect that slirp4netns terminates?

AkihiroSuda: Looks like we lack the death detection for slirp4netns too 😅 

Maybe we can use pidfd for kernel >= 5.3.
https://man7.org/linux/man-pages/man2/pidfd_open.2.html

> A PID file descriptor can be monitored using [poll(2)](https://man7.org/linux/man-pages/man2/poll.2.html), [select(2)](https://man7.org/linux/man-pages/man2/select.2.html), and [epoll(7)](https://man7.org/linux/man-pages/man7/epoll.7.html).

If we want to implement this too for old kernel, probably just watch SIGCHLD and see if the PID of `(pasta|slirp4netns)` is still valid

sbrivio-rh: > Looks like we lack the death detection for slirp4netns too 😅

I feel better now. :smile:

I think we should leave out PID tracking, because it keeps the interface to pasta very simple (also in terms of AppArmor and SELinux), and we have a valid, solid alternative (netlink monitor on the network interface). Anyway, this is the simple part.

The complicated part for me, whatever we choose as monitor, is where and how to add this to rootlesskit. I really don't know how it works with... threads? Should we spawn a thread somewhere? Is there something like that already in rootlesskit?

AkihiroSuda: We don't seem to have any monitoring stuff in the current code base, but it can be probably spawned as a goroutine

sbrivio-rh: Thanks for the tip, I'll give it a try, but with very low priority, so it might take a while.

PR #463, opened 2024-08-19
Update comment and network guide to reflect current status of pasta's AppArmor issue on Ubuntu

sbrivio-rh:
As suggested in https://github.com/rootless-containers/rootlesskit/pull/458#pullrequestreview-2242422678


AkihiroSuda (approved):
Thanks

PR #466, opened 2024-09-18
port/builtin: return proper error for sctp

Luap99:
When trying to forward sctp port we get a useless
"rootlessport spec was not validated" error back even though the spec was validated before. It is not clear to me if other drivers support sctp here so I didn't want to change ValidatePortSpec to disallow sctp.

As such simply return a useful error saying the protocol is not supported.


AkihiroSuda (approved):
Thanks

PR #482, opened 2025-01-15
pasta: set `--host-lo-to-ns-lo`; CI: update pasta

AkihiroSuda:


AkihiroSuda:
Regression in https://github.com/AkihiroSuda/passt-mirror/compare/2024_09_06.6b38f07...2024_10_30.ee7d0b6

https://github.com/rootless-containers/rootlesskit/actions/runs/12783037201/job/35633471012
> Docker integration test: net=pasta, port-driver=implicit
```
+ docker run -d --name=nginx -p 127.0.0.1:8080:80 nginx:alpine
c97d3321bb3e442209855c85ad2f348b6e9bc98dfda4a1d633c4f70896312abf
+ sleep 2
+ curl -fsSL http://127.0.0.1:8080/
curl: (56) Recv failure: Connection reset by peer
```

AkihiroSuda:
Regression in https://passt.top/passt/commit/?id=b4dace8f462b346ae2135af1f8d681a99a849a5f

"fwd: Direct inbound spliced forwards to the guest's external address"

AkihiroSuda:
Seems relevant:
- https://github.com/moby/moby/issues/48838

AkihiroSuda:
Updated PR to exec pasta with `--host-lo-to-ns-lo`, so as to keep `docker run -p 127.0.0.1:8080:80` functional with  passt >= 2024_10_30.ee7d0b6

This flag seems already deprecated, but any chance to undeprecate it?
cc @sbrivio-rh 🙏

sbrivio-rh:
> A minimal reproducer w/o Docker: `rootlesskit --net=pasta --port-driver=implicit nc -l 127.0.0.1 12345` no longer forwards the port 12345

Oops. I would argue that it's a regression for this specific test, but not for the functionality. That is, `nc -l 12345` would still work: it's just binding specifically to loopback inside the container that doesn't expose that port. If you bind to "any" address, it works.

But anyway:

> This flag seems already deprecated, but any chance to undeprecate it?

...that's something that you and probably some users expected to work, so we definitely have to keep it working.

From the related [discussion](https://archives.passt.top/passt-dev/20240925102009.62b9a0ce@elisabeth/) I would say that the only reason why it was introduced directly as deprecated is that, eventually, we would like to replace it with a more generic configuration interface (the runtime port forwarding interface I mentioned a couple of times), where one can say "map x to y" in a generic way.

But I would suggest that, until then, you can use it, and I'll make a note to find a working alternative for rootlesskit if we ever drop that option (it won't happen any time soon, anyway).

pkg/network/pasta/pasta.go
```diff
@@ -19,6 +20,35 @@ import (
...
+	if err != nil {
+		return nil, fmt.Errorf("pasta binary %q is not installed: %w", binary, err)
+	}
+	cmd := exec.Command(realBinary, "--help")
+	b, err := cmd.CombinedOutput()
+	s := string(b)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"command \"%s --help\" failed, make sure pasta is installed: %q: %w",
+			realBinary, s, err,
```
thaJeztah, line 43: Somewhat related to the "perhaps  use  `--version` suggestion (but not sure  if  it helps a lot); the failure output  may be a lot of  noise; intentionally producing an error here;

```go
package main

import (
	"fmt"
	"os/exec"
)

func main() {
	realBinary, _ := exec.LookPath("pasta")
	cmd := exec.Command(realBinary, "--heelp")
	b, err := cmd.CombinedOutput()
	s := string(b)
	if err != nil {
		e := fmt.Errorf(
			"command \"%s --help\" failed, make sure pasta is installed: %q: %w",
			realBinary, s, err,
		)
		fmt.Println(e)
	}
}
```

```bash
go run main.go
command "/usr/bin/pasta --help" failed, make sure pasta is installed: "/usr/bin/pasta: unrecognized option: heelp\nUsage: /usr/bin/pasta [OPTION]... [COMMAND] [ARGS]...\n       /usr/bin/pasta [OPTION]... PID\n       /usr/bin/pasta [OPTION]... --netns [PATH|NAME]\n\nWithout PID or --netns, run the given command or a\ndefault shell in a new network and user namespace, and\nconnect it via pasta.\n\n  -d, --debug\t\tBe verbose\n      --trace\t\tBe extra verbose, implies --debug\n  -q, --quiet\t\tDon't print informational messages\n  -f, --foreground\tDon't run in background\n    default: run in background\n  -l, --log-file PATH\tLog (only) to given file\n  --log-size BYTES\tMaximum size of log file\n    default: 1 MiB\n  --runas UID|UID:GID \tRun as given UID, GID, which can be\n    numeric, or login and group names\n    default: drop to user \"nobody\"\n  -h, --help\t\tDisplay this help message and exit\n  --version\t\tShow version and exit\n  -I, --ns-ifname NAME\tnamespace interface name\n    default: same interface name as external one\n  -F, --fd FD\t\tUse FD as pre-opened connected socket\n  -p, --pcap FILE\tLog tap-facing traffic to pcap file\n  -P, --pid FILE\tWrite own PID to the given file\n  -m, --mtu MTU\tAssign MTU via DHCP/NDP\n    a zero value disables assignment\n    default: 65520: maximum 802.3 MTU minus 802.3 header\n                    length, rounded to 32 bits (IPv4 words)\n  -a, --address ADDR\tAssign IPv4 or IPv6 address ADDR\n    can be specified zero to two times (for IPv4 and IPv6)\n    default: use addresses from interface with default route\n  -n, --netmask MASK\tAssign IPv4 MASK, dot-decimal or bits\n    default: netmask from matching address on the host\n  -M, --mac-addr ADDR\tUse source MAC address ADDR\n    default: 9a:55:9a:55:9a:55 (locally administered)\n  -g, --gateway ADDR\tPass IPv4 or IPv6 address as gateway\n    default: gateway from interface with default route\n  -i, --interface NAME\tInterface for addresses and routes\n    default: from --outbound-if4 and --outbound-if6, if any\n             otherwise interface with first default route\n  -o, --outbound ADDR\tBind to address as outbound source\n    can be specified zero to two times (for IPv4 and IPv6)\n    default: use source address from routing tables\n  --outbound-if4 NAME\tBind to outbound interface for IPv4\n    default: use interface from default route\n  --outbound-if6 NAME\tBind to outbound interface for IPv6\n    default: use interface from default route\n  -D, --dns ADDR\tUse IPv4 or IPv6 address as DNS\n    can be specified multiple times\n    a single, empty option disables DNS information\n    default: don't use any addresses\n  -S, --search LIST\tSpace-separated list, search domains\n    a single, empty option disables the DNS search list\n    default: don't use any search list\n  --dhcp-dns\t\tPass DNS list via DHCP/DHCPv6/NDP\n  --dhcp-search\tPass list via DHCP/DHCPv6/NDP\n  --map-host-loopback ADDR\tTranslate ADDR to refer to host\n    can be specified zero to two times (for IPv4 and IPv6)\n    default: gateway address\n  --map-guest-addr ADDR\tTranslate ADDR to guest's address\n    can be specified zero to two times (for IPv4 and IPv6)\n    default: none\n  --dns-forward ADDR\tForward DNS queries sent to ADDR\n    can be specified zero to two times (for IPv4 and IPv6)\n    default: don't forward DNS queries\n  --dns-host ADDR\tHost nameserver to direct queries to\n    can be specified zero to two times (for IPv4 and IPv6)\n    default: first nameserver from host's /etc/resolv.conf\n  --no-tcp\t\tDisable TCP protocol handler\n  --no-udp\t\tDisable UDP protocol handler\n  --no-icmp\t\tDisable ICMP/ICMPv6 protocol handler\n  --no-dhcp\t\tDisable DHCP server\n  --no-ndp\t\tDisable NDP responses\n  --no-dhcpv6\t\tDisable DHCPv6 server\n  --no-ra\t\tDisable router advertisements\n  --freebind\t\tBind to any address for forwarding\n  --no-map-gw\t\tDon't map gateway address to host\n  -4, --ipv4-only\tEnable IPv4 operation only\n  -6, --ipv6-only\tEnable IPv6 operation only\n  -t, --tcp-ports SPEC\tTCP port forwarding to namespace\n    can be specified multiple times\n    SPEC can be:\n      'none': don't forward any ports\n      'auto': forward all ports currently bound in namespace\n      a comma-separated list, optionally ranged with '-'\n        and optional target ports after ':', with optional\n        address specification suffixed by '/' and optional\n        interface prefixed by '%'. Examples:\n        -t 22\tForward local port 22 to port 22 in netns\n        -t 22:23\tForward local port 22 to port 23\n        -t 22,25\tForward ports 22, 25 to ports 22, 25\n        -t 22-80\tForward ports 22 to 80\n        -t 22-80:32-90\tForward ports 22 to 80 to\n\t\t\tcorresponding port numbers plus 10\n        -t 192.0.2.1/5\tBind port 5 of 192.0.2.1 to namespace\n        -t 5-25,~10-20\tForward ports 5 to 9, and 21 to 25\n        -t ~25\t\tForward all bound ports except for 25\n    default: auto\n    IPv6 bound ports are also forwarded for IPv4\n  -u, --udp-ports SPEC\tUDP port forwarding to namespace\n    SPEC is as described for TCP above\n    default: auto\n    IPv6 bound ports are also forwarded for IPv4\n    unless specified, with '-t auto', UDP ports with numbers\n    corresponding to forwarded TCP port numbers are\n    forwarded too\n  -T, --tcp-ns SPEC\tTCP port forwarding to init namespace\n    SPEC is as described above\n    default: auto\n  -U, --udp-ns SPEC\tUDP port forwarding to init namespace\n    SPEC is as described above\n    default: auto\n  --host-lo-to-ns-lo\tDEPRECATED:\n\t\t\tTranslate host-loopback forwards to\n\t\t\tnamespace loopback\n  --userns NSPATH \tTarget user namespace to join\n  --netns PATH|NAME\tTarget network namespace to join\n  --netns-only\t\tDon't join existing user namespace\n    implied if PATH or NAME are given without --userns\n  --no-netns-quit\tDon't quit if filesystem-bound target\n  \t\t\tnetwork namespace is deleted\n  --config-net\t\tConfigure tap interface in namespace\n  --no-copy-routes\tDEPRECATED:\n\t\t\tDon't copy all routes to namespace\n  --no-copy-addrs\tDEPRECATED:\n\t\t\tDon't copy all addresses to namespace\n  --ns-mac-addr ADDR\tSet MAC address on tap interface\n": exit status 1
```


Wondering if we can  come up with any situation where `--version` would work, but `--help` wouldn't (because in that case, we could potentially bail out earler?)

pkg/network/pasta/pasta.go
```diff
@@ -19,6 +20,35 @@ import (
...
+		)
+	}
+	f := Features{
+		HasHostLoToNsLo: strings.Contains(s, "--host-lo-to-ns-lo"),
```
sbrivio-rh, line 47: If it makes things simpler, you can also test the whole thing with `pasta --host-lo-to-ns-lo --version`, that is:

```
$ pasta --host-lo-to-ns-lo --version >/dev/null 2>&1 && echo option is there
option is there
$ pasta --invalid-option --version >/dev/null 2>&1 || echo option is not there
option is not there
```

By the way, sorry for breaking this. I didn't really expect users/test to actually bind *to 127.0.0.1* in the container. :(

AkihiroSuda: Thanks, updated

pkg/network/pasta/pasta.go
```diff
@@ -19,6 +20,35 @@ import (
...
+
+func DetectFeatures(binary string) (*Features, error) {
+	if binary == "" {
+		return nil, errors.New("got empty pasta binary")
+	}
+	realBinary, err := exec.LookPath(binary)
+	if err != nil {
+		return nil, fmt.Errorf("pasta binary %q is not installed: %w", binary, err)
+	}
+	cmd := exec.Command(realBinary, "--help")
```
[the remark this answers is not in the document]
AkihiroSuda: Updated to use `--version`

pkg/network/pasta/pasta.go
```diff
@@ -19,6 +20,35 @@ import (
...
+	b, err := cmd.CombinedOutput()
+	s := string(b)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"command \"%s --help\" failed, make sure pasta is installed: %q: %w",
```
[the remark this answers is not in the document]
AkihiroSuda: fixed

pkg/network/pasta/pasta.go
AkihiroSuda: Updated to just check `--version`

sbrivio-rh:
@dgibson

PR #502, opened 2025-04-01
Support IPv6 in address assignment and gateway

AlexeyPerevalov:
Rootlesskit assigns ipv4 address and routes for ipv4, but not for ipv6, this PR address IPv6 issue.

Some tools like slirp4netns could assign ip address, routes, but rootlesskit doesn't request them to do it.

Couple clues:
1. Interface address and gateway should be received from e.g. slirp4netns. There is already a comment on it in code. Need to fix slirp4netns and others.
2. Setting static IPv6 address fd00::1 to tap0 looks weired, since e.g. slirp4netns already could give dynamic address from fd00::/64 network, but there is a race. Rootlesskit can't set default route via fd00::2 unless IPv6 address and route from fd00::/64 is set to tap0.


pkg/network/parentutils/parentutils.go
```diff
@@ -6,9 +6,11 @@ import (
 	"strconv"
 
 	"github.com/rootless-containers/rootlesskit/v2/pkg/common"
+	"github.com/sirupsen/logrus"
 )
 
 func PrepareTap(childPID int, childNetNsPath string, tap string) error {
+	logrus.Debugf("PrepareTap")
```
AkihiroSuda, line 13: This debug message doesn't seem meaningful?

AkihiroSuda:
Can we have a test?

pkg/network/parentutils/parentutils.go
AlexeyPerevalov: Oh, yes. I'll remove this debug leftover.

AlexeyPerevalov:
> Can we have a test?

yes, I'll add test.

AkihiroSuda:
> > Can we have a test?
> 
> yes, I'll add test. Did you mean these tests in hack/*

Yes, thanks

pkg/network/parentutils/parentutils.go
AkihiroSuda: ping

PR #503, opened 2025-04-01
Arch in dockerfile

AlexeyPerevalov:
This PR adds parameter into Dockerfile for ISA of binaries used during image build.

The default value is x86_64 as before.
This PR tested with aarch64, and it could be specified by `--build-arg ARCH=aarch64` options of `docker build` command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,7 +77,7 @@ RUN /sbin/setcap cap_setuid+eip /usr/bin/newuidmap && \
 COPY --from=artifact /rootlesskit /home/user/bin/
 COPY --from=artifact /rootlessctl /home/user/bin/
 ARG SLIRP4NETNS_VERSION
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -6,6 +6,9 @@ ARG VPNKIT_VERSION=0.5.0
 ARG PASST_VERSION=2025_02_17.a1e48a0
 ARG DOCKER_VERSION=28.0.1
 ARG DOCKER_CHANNEL=stable
+# also tested with aarch64
+ARG ARCH=x86_64
```
AkihiroSuda, line 10: This should use `TARGETARCH` to support `docker build --platform`

https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

AkihiroSuda: Or maybe just checking `$(uname -m)` may suffice, then no need to use `ARG`

Issue #504, opened 2025-04-02
Consider vendoring dependencies


bentekkie:
Would the rootlesskit project consider checking in vendored go dependencies using https://go.dev/ref/mod#vendoring?
If so would you accept a PR to add this?

fahedouch:
> Would the rootlesskit project consider checking in vendored go dependencies using https://go.dev/ref/mod#vendoring? If so would you accept a PR to add this?

why should we enable vendoring ?

PR #505, opened 2025-04-02
Vendor go dependencies

bentekkie:
By vendoring go dependencies using https://go.dev/ref/mod#vendoring it makes it much easier to build rootlesskit in a repeatable and even offline way.

Fixes: #504

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -240,3 +240,14 @@ jobs:
         docker exec test docker info
         docker exec test ./integration-docker.sh
         docker rm -f test
+  deps:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
+    - name: install go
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
>  it much easier to build rootlesskit in a repeatable and even offline way.

I don't think this affects repeatability? (As long as github.com or proxy.golang.org is alive)

AkihiroSuda:
This increases the repo size by 12MiB, so I'd like to see a strong reason to merge this

AkihiroSuda:
An alternative idea would be to attach `vendor.tar.gz` in a release artifact

e.g., [lima-1.0.7-go-mod-vendor.tar.gz](https://github.com/lima-vm/lima/releases/download/v1.0.7/lima-1.0.7-go-mod-vendor.tar.gz) in https://github.com/lima-vm/lima/releases/tag/v1.0.7

Issue #182, opened 2020-10-22
Pdeathsig


AkihiroSuda:
> The usage of `Pdeathsig` in #66 isn't quite right. It will guarantee that the child dies, but it won't ensure that the parent doesn't accidentally kill the child. See https://github.com/golang/go/issues/27505 for details.

> The correct way to use `Pdeathsig` would be to launch a goroutine, lock it to the OS thread, start the child in that goroutine, and don't exit the goroutine until the child exits. There's a more complete example in the issue I linked above.

> FWIW, if you're launching lots of children, it might make sense to use a single goroutine locked to an OS thread to launch all of the potential children.

__Originally posted by @virtuald in https://github.com/rootless-containers/rootlesskit/issues/65#issuecomment-713709507__

PR #514, opened 2025-05-27
child: refactor command execution to use goroutines with Pdeathsig

fahedouch:
fixing https://github.com/rootless-containers/rootlesskit/issues/182

better controller for child `Pdeathsig`

### DIFF
--- a/hack/integration-pdeathsig.sh
+++ b/hack/integration-pdeathsig.sh
@@ -1,0 +1,100 @@
+#!/bin/bash
+# Test script to verify Pdeathsig behavior using rootlesskit itself
+# This script:
+# 1. Uses rootlesskit to spawn a long-running process
+# 2. Kills the rootlesskit parent process
+# 3. Verifies that the child process is killed as expected
+
+source $(realpath $(dirname $0))/common.inc.sh
+
+INFO "Starting Pdeathsig test using rootlesskit..."
+
+# Create a temporary directory for test artifacts
+TEMP_DIR=$(mktemp -d)
+INFO "Created temporary directory: $TEMP_DIR"
+
+# Create a marker file that will be touched by the child process if it's still alive
+MARKER_FILE="$TEMP_DIR/child_still_alive"
+
+# Create a script that will be executed by rootlesskit
+CHILD_SCRIPT="$TEMP_DIR/child_script.sh"
+cat > "$CHILD_SCRIPT" << 'EOF'
+#!/bin/bash
+echo "Child process started with PID: $$"
+echo "Parent PID: $PPID"
+
+# Register a trap to handle signals
+trap 'echo "Child received signal, exiting"; exit 1' TERM INT
+
+# Run for 30 seconds, checking if parent is still alive
+for i in {1..30}; do
+    echo "Child still running (iteration $i)..."
+
+    # Check if parent has changed (died)
+    CURRENT_PPID=$(ps -o ppid= -p $$)
+    if [ "$CURRENT_PPID" != "$PPID" ]; then
+        echo "Parent changed from $PPID to $CURRENT_PPID"
+        if [ "$CURRENT_PPID" = "1" ]; then
+            echo "Parent is now init (PID 1), parent has died"
+            echo "Child should be killed by Pdeathsig, but if you see this message, it wasn't"
+            touch MARKER_FILE_PLACEHOLDER
+            exit 1
+        fi
+    fi
+
+    sleep 1
+done
+
+# If we reach here, the child wasn't killed
+echo "Child completed normally (this shouldn't happen if Pdeathsig is working)"
+touch MARKER_FILE_PLACEHOLDER
+EOF
+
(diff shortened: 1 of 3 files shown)


fahedouch:
you can test with the following script:

```
@#!/bin/bash

# Test script to verify Pdeathsig behavior using rootlesskit itself
# This script:
# 1. Uses rootlesskit to spawn a long-running process
# 2. Kills the rootlesskit parent process
# 3. Verifies that the child process is killed as expected

set -e

echo "Starting Pdeathsig test using rootlesskit..."

# Create a temporary directory for test artifacts
TEMP_DIR=$(mktemp -d)
echo "Created temporary directory: $TEMP_DIR"

# Create a marker file that will be touched by the child process if it's still alive
MARKER_FILE="$TEMP_DIR/child_still_alive"

# Create a script that will be executed by rootlesskit
CHILD_SCRIPT="$TEMP_DIR/child_script.sh"
cat > "$CHILD_SCRIPT" << 'EOF'
#!/bin/bash
echo "Child process started with PID: $$"
echo "Parent PID: $PPID"

# Register a trap to handle signals
trap 'echo "Child received signal, exiting"; exit 1' TERM INT

# Run for 30 seconds, checking if parent is still alive
for i in {1..30}; do
    echo "Child still running (iteration $i)..."
    
    # Check if parent has changed (died)
    CURRENT_PPID=$(ps -o ppid= -p $$)
    if [ "$CURRENT_PPID" != "$PPID" ]; then
        echo "Parent changed from $PPID to $CURRENT_PPID"
        if [ "$CURRENT_PPID" = "1" ]; then
            echo "Parent is now init (PID 1), parent has died"
            echo "Child should be killed by Pdeathsig, but if you see this message, it wasn't"
            touch MARKER_FILE_PLACEHOLDER
            exit 1
        fi
    fi
    
    sleep 1
done

# If we reach here, the child wasn't killed
echo "Child completed normally (this shouldn't happen if Pdeathsig is working)"
touch MARKER_FILE_PLACEHOLDER
EOF

# Replace the placeholder with the actual marker file path
sed -i "s|MARKER_FILE_PLACEHOLDER|$MARKER_FILE|g" "$CHILD_SCRIPT"
chmod +x "$CHILD_SCRIPT"

# Start rootlesskit with the child script
echo "Starting rootlesskit..."
rootlesskit "$CHILD_SCRIPT" &
ROOTLESSKIT_PID=$!
echo "Rootlesskit started with PID: $ROOTLESSKIT_PID"

# Wait a moment for the child to start
sleep 2

# Find the child process
CHILD_PID=$(pgrep -P $ROOTLESSKIT_PID)
if [ -z "$CHILD_PID" ]; then
    echo "Failed to find child process"
    exit 1
fi
echo "Found child process with PID: $CHILD_PID"

# Kill the rootlesskit process
echo "Killing rootlesskit process (PID: $ROOTLESSKIT_PID)..."
kill -9 $ROOTLESSKIT_PID

# Wait a moment for the child to be killed
sleep 2

# Check if the child process is still running
if ps -p $CHILD_PID > /dev/null; then
    echo "FAIL: Child process (PID: $CHILD_PID) is still running after parent was killed"
    kill -9 $CHILD_PID  # Clean up
    exit 1
else
    echo "PASS: Child process (PID: $CHILD_PID) was killed as expected"
fi

# Check if the marker file exists
if [ -f "$MARKER_FILE" ]; then
    echo "FAIL: Marker file exists, which means the child process wasn't killed by Pdeathsig"
    exit 1
else
    echo "PASS: Marker file doesn't exist, which means the child process was killed by Pdeathsig"
fi

echo "Test completed successfully!"
rm -rf "$TEMP_DIR"
exit 0
````

I can convert this script into integration test if needed.

**Comment by AkihiroSuda** [comment]
> integration test

Yes, please
https://github.com/rootless-containers/rootlesskit/blob/e83d7635183e1125798b2928b22002dfcc4a1168/.github/workflows/main.yaml#L40-L54

File: hack/integration-pdeathsig.sh [hunk 1]
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
AkihiroSuda, line 1: Needs set -eu -o pipefail

fahedouch:
The latest commit introduces the following changes:

1. Modified `child.go` to ensure consistent Pdeathsig behavior regardless of reaper setting:
   - Added a goroutine with `runtime.LockOSThread()` for the non-reaper case (`opt.Reaper` is false)
   - Explicitly set Pdeathsig with `unix.Prctl(unix.PR_SET_PDEATHSIG, uintptr(unix.SIGKILL), 0, 0, 0)` in both cases
   - Created a new `runWithoutReap()` function to handle command execution without reaping

2. Enhanced `integration-pdeathsig.sh` to:
   - Test both reaper settings (true and false) to verify consistent behavior

new changes ensure that child processes are properly terminated when the parent is killed with SIGKILL, regardless of whether the reaper functionality is enabled.

test fail not related to PR

PR #607, opened 2026-07-09
api: expose IPv6 capability in NetworkDriverInfo

Akshitguptaa:
Adds an `IPv6` bool to `NetworkDriverInfo` in the API. This gives consumers (like nerdctl) a proper way to check for IPv6 support instead of parsing CLI help text.

- Added `IPv6` field to `api.go` and `openapi.yaml`
- Updated `gvisortapvsock`, `pasta`, and `slirp4netns` drivers to expose their `enableIPv6` state

Ref: https://github.com/containerd/nerdctl/pull/5055#discussion_r3548420865

PR #613, opened 2026-07-23
port/builtin: switch from iptables to nft for transparent routing

### LINKED ISSUE #589: port/builtin: switch away from iptables to nft
https://github.com/rootless-containers/rootlesskit/blob/be156df0429551f5c23dc61dd07201b33076ff1d/pkg/port/builtin/child/child.go#L215-L249

Akshitguptaa:
Fixes #589

### DIFF
--- a/pkg/port/builtin/child/child.go
+++ b/pkg/port/builtin/child/child.go
@@ -1,6 +1,7 @@
 package child
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -22,6 +23,10 @@ import (
 	opaquepkg "github.com/rootless-containers/rootlesskit/v3/pkg/port/builtin/opaque"
 )
 
+// probeTimeout bounds every backend-probe subprocess so a stuck or slow
+// nft/iptables invocation can never hang the init handshake indefinitely.
+const probeTimeout = 3 * time.Second
+
 func NewDriver(logWriter io.Writer) port.ChildDriver {
 	return &childDriver{
 		logWriter: logWriter,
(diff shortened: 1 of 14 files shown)


pkg/port/builtin/child/child.go
AkihiroSuda, line 1: Let's keep iptables mode too and test the both in CI

Akshitguptaa: kept iptables as a fallback.
CI runs test with both nft and iptables, and a separate one with only iptables (no nft) to cover the fallback path.

pkg/port/builtin/child/child.go
```diff
@@ -207,6 +207,9 @@ fallback:
 
 // setupTransparentRouting sets up policy routing so that response packets
 // destined to transparent-bound source IPs are delivered locally.
+// The firewall rules are implemented via nft, falling back to iptables if
+// nft isn't available on the host (see setupTransparentRoutingNFT and
+// setupTransparentRoutingIPTables).
```
AkihiroSuda, line 296: Can we have a CLI flag `--source-ip-transparent-backend` to explicitly set the backend?

Also, the backend should be visible in `PortDriverInfo`, perhaps as `Extra map[string]string`

See https://github.com/rootless-containers/rootlesskit/pull/607 for how to extend the API

AkihiroSuda (approved):
Thanks, looks good, but let me check this again by the end of the month

Akshitguptaa:
> Thanks, looks good, but let me check this again by the end of the month

One thing worth a look before then: `PortDriverInfo.Extra` currently reports the *configured* `--source-ip-transparent-backend`, not the runtime-selected one. Backend selection happens lazily on the first transparent connection, so `Extra` can say "auto" while the child is actually running iptables (or hasn't decided anything yet).

Two ways to fix it:

1. make routing setup eager (run at init, not lazily on first  connection) - smaller diff, but changes existing behavior: firewall rules get installed at startup even if source-ip-transparent is never actually used

2. keep it lazy, make Info() a live on-demand query to the child - bigger diff, no behavior change

Which do you prefer?

AkihiroSuda:
Thanks, 1 seems better

AkihiroSuda:
CI keeps failing
https://github.com/rootless-containers/rootlesskit/actions/runs/31284105470/job/97921126958?pr=613

Also needs rebasing

Akshitguptaa:
PTAL @AkihiroSuda

AkihiroSuda (approved):
Thanks

PR #616, opened 2026-08-01
feat: support --port-driver=pesto for IPv4

### LINKED ISSUE #611: Support port management for --net=pasta via pesto
## Suggestion/Question

`pesto`, a recently added command, can dynamically expose/unexpose ports on a running pasta instance.

Details of pesto:

```
NAME
       pesto - Configure a running passt(1) or pasta(1) instance.

SYNOPSIS
       pesto [OPTION]... PATH

DESCRIPTION
       pesto is a client to view and update the port forwarding configuration of a
       running passt(1) or pasta(1) instance.

       PATH  gives  the  path to the UNIX domain socket created by passt or pasta.
       It should match the -c command line option given to that instance.
```

This feature did not exist when the pasta network driver was first added to rootlesskit. It became available only recently.

Specifically, `pesto -A` and `pesto -D` allow exposing and unexposing ports respectively.

I would like to implement a new port driver based on pesto so that `rootlessctl (list-ports|add-ports|remove-ports)` works with `--net=pasta`. Would this approach be acceptable?

Note that when rootlesskit is run with `--net=pasta`, the following pasta command is executed internally:

```bash
pasta --stderr --ns-ifname=tap0 --mtu=65520 --config-net --address=10.0.2.100 --netmask=24 --gateway=10.0.2.2 --dns-forward=10.0.2.3 --no-map-gw --ipv4-only --tcp-ports=auto --udp-ports=auto
```

With `--tcp-ports=auto --udp-ports=auto`, every port that is listened on inside the namespace created by pasta is exposed to the host indiscriminately, which can lead to unintended exposure. In the context of rootless containers, I think it is safer to expose only explicitly specified ports.

---
# Background

The following demonstrates exposing and unexposing ports in the namespace created by pasta. 
Note that this feature requires the UNIX domain socket created by pasta's `-c` option.

<details><summary>Details</summary>
<p>

On Terminal 1:

Start pasta with `-t none -u none` so that no ports are forwarded.

```
(host) $ pasta --config-net -c /tmp/conf.sock -t none -u none --trace -l /tmp/pasta.log -- python3 -m http.server 8080 
...
Serving HTTP on 0.0.0.0 port 8080 (http://0.0.0.0:8080/) ...

```

On Terminal 2 (from host ns):

From the host namespace, we cannot reach the server yet, because no ports are exposed.

```bash
(host) $ curl -4 localhost:8080
curl: (7) Failed to connect to localhost port 8080 after 0 ms: Could not connect to server
...
```

However, once port 8080 is exposed using pesto, we can reach the server from the host namespace.

```
(host) $ pesto -s -A -t 8080 /tmp/conf.sock
Previous configuration (/tmp/conf.sock)
  HOST
  SPLICE
Updated configuration (/tmp/conf.sock)
  HOST
    TCP [*]:8080  =>  8080
  SPLICE
```

```bash
(host) $ curl -4 localhost:8080 2> /dev/null | head -3
<!DOCTYPE HTML>
<html lang="en">
<head>
```

The exposed port can then be unexposed with the following command.

```
(host) $ pesto -s -D -t 8080 /tmp/conf.sock
Previous configuration (/tmp/conf.sock)
  HOST
    TCP [*]:8080  =>  8080
  SPLICE
Updated configuration (/tmp/conf.sock)
  HOST
  SPLICE
```

As a result, the server running in the isolated namespace can no longer be reached from the host namespace.

```
(host) $ curl -4 localhost:8080
curl: (7) Failed to connect to localhost port 8080 after 0 ms: Could not connect to server
```

</p>
</details>

haytok:
related: https://github.com/rootless-containers/rootlesskit/issues/611


haytok:
Checking the reason of failed tests ...

haytok:
ready for review.

AkihiroSuda (approved):
Thanks

PR #620, opened 2026-08-24
fix(pasta): assign a fixed IPv6 address, gateway, and DNS to the netns

haytok:
Currently, `curl -fsSL http://[${parent_ipv6}]:8080` fails when running `hack/integration-ipv6.sh` with `--net=pasta`.

This is because pasta copies the destination address `${parent_ipv6}` into the isolated netns and the destination becomes a local address inside the netns.

To begin with, the only interface on the host with an available IPv6 route is the dummy interface (`dummy42`), so it is selected as the template, and its address and route are duplicated into the netns by `nl_addr_dup()` and `nl_route_dup()` [1].

As a result, the route to `${parent_ipv6}` becomes `local`, the packets never leave the netns, and they do not reach the httpd running on the host.

To resolve this error, pass `--address`, `--gateway`, and `--dns-forward` when running pasta, so that a fixed IPv6 configuration is assigned to tap0 inside the netns.

  child:   fd00::100
  gateway: fd00::2
  DNS:     fd00::3

With these options, pasta executes `nl_addr_set()` and `nl_route_set_def()` [1] in `pasta.c` instead of `nl_addr_dup()` and `nl_route_dup()`, and assigns the specified address and route to tap0 inside the netns.

As a result, `${parent_ipv6}` is no longer duplicated into the netns, and when running `curl -fsSL http://[${parent_ipv6}]:8080` inside the netns, packets leave the netns through tap0.

On the host side, pasta relays the connection to the httpd, so the curl now succeeds.

[1] https://passt.top/passt/tree/pasta.c

### DIFF
--- a/pkg/network/iputils/iputils.go
+++ b/pkg/network/iputils/iputils.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"math"
+	"math/big"
 	"net"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/network/iputils/iputils.go
```diff
@@ -21,3 +22,18 @@ func AddIPInt(ip net.IP, i int) (net.IP, error) {
 	binary.BigEndian.PutUint32(res, uint32(resInt64))
 	return res, nil
 }
+
+func AddIPInt6(ip net.IP, i int) (net.IP, error) {
```
AkihiroSuda, line 26: Instead you could call `https://pkg.go.dev/net/netip#Addr.Next` multiple times ?

AkihiroSuda (approved):
Merging anyway, thanks

pkg/network/iputils/iputils.go
haytok: Thanks, will fix on another PR.

haytok:
Thanks for your review!!

By the way, what would it take for `--net=pasta` and `--port-driver=pesto` to no longer be marked as experimental?

AkihiroSuda:
> By the way, what would it take for `--net=pasta` and `--port-driver=pesto` to no longer be marked as experimental?

Let's wait for a couple of months to confirm that nobody is experiencing a serious issue

PR #621, opened 2026-08-24
chore: apply go fmt ./...

haytok:


AkihiroSuda (approved):
Thanks, we have to enforce this in CI

PR #625, opened 2026-08-26
ci: enforce go fmt ./... in CI

### LINKED ISSUE #624: CI does not enforce Go code formatting
CI currently has no step that checks Go formatting 

this came up in #621 , where @AkihiroSuda commented

> Thanks, we have to enforce this in CI

Two options:  
1. Add a simple `gofmt -l` check to CI
2. Adopt `golangci-lint` with formatting enforcement built in (like nerdctl does)

haytok:
related: https://github.com/rootless-containers/rootlesskit/pull/621#pullrequestreview-5022504135

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,6 +1,20 @@
 name: Main
 on: [push, pull_request]
 jobs:
+  lint:
+    name: "Lint"
+    runs-on: ubuntu-24.04
+    steps:
+    - name: "Check out"
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+    - name: "Set up Go"
+      uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e  # v7.0.0
+      with:
+        go-version-file: go.mod
+    - name: "golangci-lint"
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -1,6 +1,20 @@
...
+    - name: "Check out"
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+    - name: "Set up Go"
+      uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e  # v7.0.0
+      with:
+        go-version-file: go.mod
+    - name: "go fmt"
+      run: |
+        go fmt ./...
+        git diff --exit-code
```
AkihiroSuda: Any reason not to use golangci-lint?

haytok: There's no specific reason. Other projects (e.g., nerdctl and Lima) use golangci-lint, so I've introduced it here as well. Thanks.

haytok:
Note that I'll address TODOs in `.golangci.yml` later via a follow-up PR since fixing them all here would make the diff too large.

AkihiroSuda (approved):
Thanks

PR #640, opened 2026-09-13
refactor: use `netip.Addr.Next()` instead of `math/big` in `AddIPInt6`

haytok:
follow-up: https://github.com/rootless-containers/rootlesskit/pull/620#discussion_r3957446613


AkihiroSuda (approved):
Thanks

PR #642, opened 2026-09-21
fix(pasta): remove --stderr option

haytok:
According to the passt commit ([1]) and source code ([2]), --stderr does nothing and is deprecated.

Therefore, this commit removes the option.

[1] https://passt.top/passt/commit/?id=bca0fefa32e083ab46a49989f225ec34e7cdbb8c
[2] https://passt.top/passt/tree/conf.c?id=588b545dae741bec6fd7622a33c7852c06d72a59#n1575


AkihiroSuda (approved):
Thanks

